PR #2, opened 2017-11-08
SJRK-16: migrating waharnum repo to fluid-project.

BlueSlug:
To be reviewed by @simonbates


Gruntfile.js
```diff
@@ -0,0 +1,36 @@
...
+
+module.exports = function (grunt) {
+    "use strict";
+
+    // Project configuration.
+    grunt.initConfig({
+        // Project package file destination.
+        pkg: grunt.file.readJSON("package.json"),
+        eslint: {
+            all: ["src/**/*.js", "tests/**/*.js", "demos/**/*.js", "examples/**/*.js", "*.js"]
```
simonbates: We don't need the entries here for `"demos/**/*.js"` and `"examples/**/*.js"`.

Gruntfile.js
```diff
@@ -0,0 +1,36 @@
...
+        },
+        jsonlint: {
+            all: ["package.json", ".jshintrc", "src/**/*.json", "tests/**/*.json", "demos/**/*.json", "!node_modules", "!src/lib/**", "!tests/lib/**"]
```
simonbates: `".jshintrc"` should be `".eslintrc.json"`.

And we can remove these entries: `"demos/**/*.json"`, `"!node_modules"`, `"!src/lib/**"`, and `"!tests/lib/**"`.

README.md
```diff
@@ -1 +1,5 @@
-Placeholder README for migration of personal repos.
+# sjrk-storyTelling
+
+SJRK Storytelling Project.
+
+UI (so far) for storytelling project.
```
simonbates: Could we expand this README?

Some thoughts:

- Contextual information, such as brief project background and links to project website (if applicable)
- Overview of what is contained in this repo
- Any instructions to install/run/test what is here

package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "sjrk-storytelling",
```
simonbates: We have called the repo "sjrk-story-telling".

package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "0.0.1",
+  "description": "SJRK Storytelling Project",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
simonbates: Looks like this project's tests are all browser-based, so we can remove this `"test"` script command.

src/js/templatedComponent.js
```diff
@@ -0,0 +1,201 @@
...
+
+/* global fluid, sjrk */
+
+(function ($, fluid) {
+
+    "use strict";
+
+    // Component that renders its UI using an external
+    // HTML file with fluid.stringTemplate syntax
+    // see concrete components in storyTelling.js for examples
```
simonbates: We don't have a file called `storyTelling.js`

src/js/storyTelling-story.js
```diff
@@ -0,0 +1,100 @@
...
+            translationOf: null
+        }
+    });
+
+    // mixin grade which combines the base story with a templatedComponentWithLocalization
+    // contains common elements for all of the story interface contexts (editor, viewer, etc.)
+    fluid.defaults("sjrk.storyTelling.story.ui", {
+        gradeNames: ["sjrk.storyTelling.story.base", "sjrk.storyTelling.templatedComponentWithLocalization"],
+        model: {
+            templateTerms: {
```
simonbates: I'm seeing 3 types of things in models of localized view components:

- story data, such as `model.title`, `model.content`
- machine content such as ids (`model.templateTerms.storyTitleIdForLabel`) and classes (`model.templateTerms.storyTitleClasses`)
- user interface text strings (as loaded from `storyMessages_??.json` into `model.templateTerms` by `sjrk.storyTelling.messageLoaderWithLocalization.loadLocalizationMessages()`)

In addition, we have some duplication, such as `model.templateTerms.storyTitle: "{that}.model.title"`.

I think that the design would be clearer if we were to refactor so that the component `model` contained only story data. And that the machine ids and UI text strings could be moved out of the model to another location on the component.

For example, in the Infusion preferences framework, we use `options.strings` for UI text strings:

https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/Panels.js#L673

With a `fluid.prefs.msgLookup` grade:

https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/Panels.js#L23-L34

src/js/templatedComponent.js
```diff
@@ -0,0 +1,201 @@
...
+            // The template file using the
+            // fluid.stringTemplate syntax
+            // componentTemplate: ""
+        },
+        terms: {
+            "resourcePrefix": "."
+        }
+    });
+
+    fluid.defaults("sjrk.storyTelling.messageLoaderWithLocalization", {
```
simonbates: I think that it would be worthwhile to do some refactoring around the specification of component templates and localized messages. We have a few different forms of inheritance/merging going on here and I think we can simplify.

We have 2 grades for specifying component templates:

- `sjrk.storyTelling.messageLoader`
- `sjrk.storyTelling.messageLoaderWithLocalization` extends `messageLoader`, adds `locale` and `defaultLocale` of `"en"`

And 2 related components that use templates:

- `sjrk.storyTelling.templatedComponent` has a `messageLoader` called `resourceLoader`
- `sjrk.storyTelling.templatedComponentWithLocalization` extends `templatedComponent`, replacing `resourceLoader` with a `messageLoaderWithLocalization`

We also have a hierarchy with:

- `sjrk.storyTelling.story.ui` is a `sjrk.storyTelling.templatedComponentWithLocalization`
    - specifies a `componentMessages` resource in its `resourceLoader`
- leaf UI components that use `sjrk.storyTelling.story.ui` (such as `sjrk.storyTelling.story.storyViewer`)
    - each argument the `resourceLoader` with a `componentTemplate`

One of the consequences of this design is that the `componentTemplate` resources of `sjrk.storyTelling.story.ui` components will be loaded using the localized search from `sjrk.storyTelling.messageLoaderWithLocalization` (because the `resourceLoader` is of type `messageLoaderWithLocalization`). It's not clear to me that this is intentional. The templates for `sjrk.storyTelling.story.ui` components are located with `storyAuthoring.handlebars`, which does not use the localized loading. The template specification looks the same but are actually different (one uses the localized logic and the other does not).

From `sjrk.storyTelling.story.storyViewer`:

    resourceLoader: {
        options: {
            resources: {
                componentTemplate: "%resourcePrefix/src/templates/storyView.handlebars"
            }
        }
    }

And from `sjrk.storyTelling.storyAuthoring`:

    resourceLoader: {
        options: {
            resources: {
                componentTemplate: "%resourcePrefix/src/templates/storyAuthoring.handlebars"
            }
        }
    }

The structure may be clearer with more use of component composition over grade merging. In particular, I suggest using separate resource loaders for templates and messages -- one with no locale and one with locale.

src/js/templatedComponent.js
```diff
@@ -0,0 +1,201 @@
...
+        resources: {
+            // Specified by using grade
+            // The messages file (JSON)
+            // componentMessages: ""
+        },
+        locale: "en",
+        defaultLocale: "en"
+    });
+
+    sjrk.storyTelling.messageLoaderWithLocalization.loadLocalizationMessages = function (componentMessages, that, modelEndpoint) {
```
simonbates: We should use the component model for mutable user data only and store the user interface messages elsewhere.

src/js/templatedComponent.js
simonbates: To add to and correct what I wrote above: the `storyAuthoring.handlebars` file is actually loaded with localization search. While it is loaded with a `sjrk.storyTelling.messageLoader` (rather than `messageLoaderWithLocalization`), the `sjrk.storyTelling.storyAuthoring` grade has a `distributeOptions` that applies a `locale`:

    resourceLoaderConfig: {
        resourcePrefix: ".",
        locale: "en"
    },
    distributeOptions: [{
        source: "{that}.options.resourceLoaderConfig.resourcePrefix",
        target: "{that resourceLoader}.options.terms.resourcePrefix"
    },
    {
        source: "{that}.options.resourceLoaderConfig.locale",
        target: "{that resourceLoader}.options.locale"
    }],

If I load the top level `index.html` in Firefox and look at the Network developer tab, I see:

- 200 `/src/templates/storyAuthoring.handlebars`
- 404 `/src/templates/storyAuthoring_en.handlebars`
- 200 `/src/templates/storyEdit.handlebars`
- 404 `/src/messages/storyMessages.json`
- 404 `/src/templates/storyEdit_en.handlebars`
- 200 `/src/messages/storyMessages_en.json`

BlueSlug:
@michelled and I have been considering the template files (*.handlebars) and the way they have been using the CSS classes and DOM element ID's defined in the templateTerms blocks.
I don't think it's necessary at this point to make any changes to them. The reasoning being that there are very likely new radically different designs on the way, and any work done at this point may effectively be thrown out in the near future anyway.

src/js/transforms.js
```diff
@@ -0,0 +1,51 @@
...
+
+/* global fluid, sjrk */
+
+(function ($, fluid) {
+
+    "use strict";
+
+    fluid.registerNamespace("sjrk.storyTelling.transforms");
+
+    fluid.defaults("sjrk.storyTelling.transforms.stringToArray", {
```
michelled: Is it worth contributing these to Infusion? They seem generally useful.

BlueSlug: I think they're quite useful, and as you've suggested outside of this discussion, there doesn't seem to be an existing equivalent in Infusion already.

The closest I was able to find in the existing transforms is [`fluid.transforms.JSONstringToObject`](https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/ModelTransformationTransforms.js#L809-L822) and its inverse function, [`fluid.transforms.objectToJSONString`](https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/ModelTransformationTransforms.js#L834-L857), but these deal with JSON objects or serializations, and in this case we'd like to have it produce arrays, specifically.

A follow-up question: is it worth perhaps using the above-mentioned transform instead of the one we've implemented in `transforms.js`?

BlueSlug:
After discussions with @colinbdclark, it would be easier at this point to consider this branch stale and make a fresh PR from SJRK-91. Therefore, I have created #3 and am closing this PR

PR #5, opened 2019-01-18
SJRK-174: merging in latest changes before other work

BlueSlug:
- Added contact email address
- New icons for the editor
- Minor styling updates
- Updated a few messages


simonbates:
@BlueSlug Does this PR require any updates to sjrk-story-telling-server?

BlueSlug:
@simonbates I am currently in the process of testing this, but my initial thoughts are that it does not. Once I finish my testing on the server project, I'll make an update in this thread.

BlueSlug:
@simonbates based on my local testing (after my most recent commit to fix another issue), there are no changes required to the server project to accommodate the changes in this PR.

simonbates:
The video block has a field "Translate the audio to text (optional)". Should that be "video"?

BlueSlug:
@simonbates this is something I was unsure about while filling in that message value. Considering the transcript is meant to be a textual representation of the audio portion of the file, I chose to go with "audio" rather than "video", but I can see a case for both. Ultimately I chose the former, but I have no problems changing it to "video". What do you think?

Perhaps there's a case to be made for something more obvious, like "Translate the video's audio to text (optional)", though that is a bit clunky.

PR #11, opened 2019-01-28
SJRK-207: Add additional "clicked" state to block editor buttons

BlueSlug:
This requires no changes to the server project


simonbates:
I believe that we have some new icons added that aren't used in the current code:

- `Forward-disabled.svg`
- `Trashcan-disabled.svg`

Is the intention that these will be used in future work?

simonbates:
You have renamed `placeholder-image.svg` to `Text-background.svg`. Looking at the image, I think it should be `Image-background.svg`.

**Update:** I see that [SJRK-195](https://github.com/fluid-project/sjrk-story-telling/pull/12) renames `placeholder-image.svg` to `Image-background.svg` and will conflict with this one.

simonbates:
It looks like the `*-background.svg` images aren't currently used? This will be done in future work?

**Update:** I see that [SJRK-195](https://github.com/fluid-project/sjrk-story-telling/pull/12) uses the background images. It also renames them, and will therefore conflict with this PR.

simonbates:
I see that this isn't new to this pull request but Audio/Voice naming is confusing. The icons for the other block types (Text, Image, Video) follow the media type. Whereas the icon for the audio block is "Voice". And "Audio" is used for the text-to-speech self-voicing. Could we switch the names Audio <-> Voice?

BlueSlug:
re the `*-disabled.svg`:
- `Trashcan-disabled.svg` will be used according to the work described in https://issues.fluidproject.org/browse/SJRK-193
- `Forward-disabled.svg` may not end up being used, as it turns out

BlueSlug:
re Audio/Voice naming confusion, I agree 100%. The naming convention I have used here lines up with the names of the asset bundles I receive from @sepidehshahi, so perhaps this is something the three of us should discuss when we have some time.

cindyli:
This pull request works well with mouse clicking on buttons, but there seems an issue with having the active state shown when keyboard activating these buttons. To reproduce:

1. Use "tab" key to focus on "adding a text block" button, or any other media buttons;
2. Hit "Enter" or "Space" to insert a block;
3. Expected: with Chrome and Safari, hitting "Space" inserts the block and the button active state shows;
4. Unexpected: with Chrome, Safari and Firefox, hitting "Enter" inserts the block but doesn't show the active state. With Firefox, hitting "Space" doesn't show the active state too.

These browsers are tested on Mac.

BlueSlug:
I was able to confirm this issue in Firefox on Windows as well. It seems to behave as expected in Chrome and Edge.

Re the Enter key, the behaviour for that seems to be repeatedly pressing the button anyway, so this probably should not be expected in any case.

Testing this page from W3Schools in Firefox on Windows, I was able to see that even an unstyled `button` does not have an `:active` state when the spacebar is pressed:
https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_button_test

I've discussed the issue with @sepidehshahi, and while this is obviously not ideal, what we have in this PR so far is better than before (which is no indication at all).

I think we have two options at this point:
1. Leave this change as-is and gain the functionality using the mouse or using spacebar in certain browsers
2. I can try to implement a solution along the lines of using jQuery to inject a CSS rule on `keyup` / `keydown` for space and enter

BlueSlug:
@cindyli After discussing the matter with @sepidehshahi, we should proceed with your plan to merge this current work and file a new Jira to track the "bug" in Firefox where the spacebar doesn't lead to the `:active` state being applied.

BlueSlug:
SJRK-225 has been filed to track this issue
https://issues.fluidproject.org/browse/SJRK-225

cindyli:
Merged at 856b5a93ebc63d13706920bfb57446353c792faf

PR #12, opened 2019-01-29
SJRK-195: Add a background image to each block editor

BlueSlug:
This update requires no changes to the server project.
 
Adds backgrounds for media block types when there is no file uploaded. This does not apply to text blocks as no asset for that was provided, nor is there a button to "activate" the text block with.

To note: the Chrome user agent stylesheet has a rule which hides audio elements without controls, and this cannot be overridden regardless of use of `!important` rules and, frustratingly, completely ignoring CSS specificity rules. To get around this, I've added an empty div which, using the CSS adjacent combinator (`+`), will make it appear with the correct styling only when its audio element sibling has an empty or missing `src` attribute.

This has been tested on Windows 10 in Chrome 71, Firefox 64 and Edge 44.


src/js/storyTelling-blockUi-timeBased.js
```diff
@@ -89,12 +89,14 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         // for the source element of the mediaPlayer.
         // Alternatively, we could go with something simpler based on directly using the
         // src attribute of the video or audio element.
-        var source = mediaPlayer.find("source");
-        if (source.length < 1) {
-            source = $("<source></source>").appendTo(mediaPlayer);
+
+        if (mediaUrl) {
```
cindyli: This if block can be simplified to: `mediaPlayer.prop("controls", !!mediaUrl);`

BlueSlug: I've never seen that "operator" before! very neat. thanks for the tip :) I'll make that change once I return to working on this branch

cindyli:
A couple of UI issues:

1. When adding a text block, the background image is not shown:
screen shot 2019-02-11 at 10 26 05 am

2. When adding an image block, the background image shows with a black border around it with firefox and safari on Mac. Chrome on Mac shows the background image properly without a black border:
screen shot 2019-02-11 at 10 26 20 am

BlueSlug:
@cindyli thanks for you feedback. For the text block background, I didn't receive an image for that, and in the case where there is no text it's just an empty text box. I'll grand that t is a bit inconsistent.

I've tried to find a solution to the border around the image box, but I haven't come up with one yet. The issue seems to stem from the fact that the image has a `width` value set in CSS. If it weren't for that, the border would not appear. It is not solvable using the `border` CSS property either.

BlueSlug:
@cindyli After following up with @sepidehshahi, the border around the image box is acceptable as it is. For the time being, we will direct Mac users in story jams to use Chrome to circumvent the issue, though I'll make a minor change to make these backgrounds more in line with the design across all browsers anyway. Text blocks are fine as-is without the background.

We should consider tracking this CSS bug in a separate Jira, as we're doing with #11

BlueSlug:
SJRK-226 has been filed to track the styling bug
https://issues.fluidproject.org/browse/SJRK-226

cindyli:
Merged at 5d27c25f00ec2a996b6ca233e73d89cfc3df6a45

PR #16, opened 2019-02-14
SJRK-216: Add progress indicator on story submission

BlueSlug:
This requires changes to the server project as well: https://github.com/fluid-project/sjrk-story-telling-server/pull/14

Adds a progress indicator that is shown upon hitting the "publish" button. The button itself is disabled while the site is making a request to the server. The request will result in either the user being redirected to their new story (success) or the progress indicator being removed, the button being re-enabled and an error message being shown to indicate the publishing has failed.

This also moves the events related to clicking the publish button up into the Edit page so that the server project isn't reaching into that component's subcomponents.


src/js/storyTelling-page-storyEdit.js
```diff
@@ -93,8 +95,77 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                            args: [false]
+                        },
+                        disableShareButton: {
+                            func: "{that}.setShareButtonDisabled",
+                            args: [true]
+                        },
+                        showProgressArea: {
+                            this: "{that}.dom.progressArea",
+                            method: "show",
+                            args: [0]
```
cindyli, line 147: Why `args: [0]`? The same comment applies to other invokers below.

src/templates/storyViewer.handlebars
```diff
@@ -46,7 +46,7 @@
     <div lang="{{localizedMessages.language}}" class="sjrk-st-story-viewer-footer-container">
         <div>
             <a href="javascript:void(0)" class="sjrkc-st-story-viewer-previous sjrk-st-story-viewer-previous">
-                <img src="/src/img/icons/icon-back.svg" alt="left arrow icon">
+                <img src="/src/img/icons/icon-back.svg" alt="{{localizedMessages.message_goBackIcon}}" aria-hidden="true">
```
cindyli, line 49: The value of "aria-hidden" should be flipped programmatically.

cindyli: The same comment applies to the similar changes in `src/templates/storyEditor.handlebars`.

tests/js/storyTelling-page-storyEditTests.js
```diff
@@ -67,6 +67,24 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         }
     });
 
+    var expectedVisibility = {
+        prePublish: {
+            progressArea: false,
+            responseArea: false,
+            shareButton: false
```
cindyli, line 74: Why is the share button invisible before and after the publish?

cindyli: I see, it's the "disable" state rather than the visibility for the share button. Please ignore this.

src/templates/storyViewer.handlebars
```diff
@@ -57,5 +57,12 @@
         <div hidden>
             <a href="#" class="sjrkc-st-story-save-no-share sjrk-st-story-save-no-share">{{localizedMessages.message_saveNoShareText}}</a>
         </div>
+        <div class="sjrkc-st-story-share-progress sjrk-st-story-share-progress" aria-live="assertive">
+            <div class="sjrk-st-story-share-progress-indicator" aria-labelledby="sjrkc-st-story-share-progress-text"></div>
```
cindyli, line 61: As discussed with @BlueSlug and @klown at fluid-work channel (the discussion can be found found in [this log](http://irc-logs.fluidproject.org/%23fluid-work/%23fluid-work.2019-02-15.log) from 2019-02-15T21:46:22), @BlueSlug will look into using [aria progressbar attritute](https://www.w3.org/TR/wai-aria-1.1/#progressbar) for this progress indicator.

cindyli:
When the connection to the database is broken, the publishing progress indicator will show for a couple of minutes to eventually display the server returned error. This seems a long wait. I wonder if you could discuss with a designer to find out if it's necessary to implement a client side timeout mechanism to reduce the wait time.

src/js/storyTelling-page-storyEdit.js
BlueSlug: According to the docs for the jQuery `show` and `hide` methods, the default is to apply an animation to the operation. I don't want the animation to show, so I'm passing in a value of `0` to the first argument, `duration`.

documentation here: http://api.jquery.com/show/

cindyli:
Please merge in the upstream to avoid merge conflicts. Thanks.

src/js/storyTelling-page-storyEdit.js
cindyli: I see. Thanks for the explanation.

BlueSlug:
@cindyli In investigating the issue of a broken connection to the database and related two-minute timeout from the `$.ajax()` call, this is a result of the issue described in [SJRK-227](https://issues.fluidproject.org/browse/SJRK-227). Do you think this should be addressed as a part of SJRK-216?

cindyli:
As long as this long-wait issue is tracked and on the plan, it doen't need to be addressed here.

cindyli:
Running `html/storyTelling-page-Tests.html` fails with this error: 
screen shot 2019-02-19 at 1 56 13 pm

Similar errors with running `html/storyTelling-page-storyEdit-Tests.html` and `html/storyTelling-page-storyBrowse-Tests.html`.

src/templates/storyViewer.handlebars
BlueSlug: My intention was to hide the additional info in the accessibility tree related to this link, as any alt-text for this image would be identical to the text displayed or would be otherwise redundant, superfluous or even potentially distracting or confusing. I.e. if the alt-text value for the image is "left arrow icon", NVDA says "left arrow icon graphic back link" rather than simply "back link"

BlueSlug: Reviewing the [WAI-ARIA recommendations](https://www.w3.org/TR/wai-aria-1.1/#aria-hidden), it cautions against using `aria-hidden` in this way without first considering the accessibility ramifications of the decision, but I believe this is a change for the better. I can certainly change it back if you feel it isn't appropriate.

BlueSlug:
@cindyli I believe your issues with testing stem from mounting the site in a subdirectory of your local server. I have noticed this issue in the past and have now filed a Jira to address the matter: https://issues.fluidproject.org/browse/SJRK-232

That said, when I tested the site out in a subfolder I was seeing many more failing tests than just the `page` component and its derivatives.

It would be helpful to know what software you're using to run your development server! :)

cindyli:
Yes, it's true that SJRK is served from a subdirectory of my local server. I use MAMP package that uses Apache to host my local server. I have a "Development" directory that has all projects as subdirs and the Apache document root dir is pointing to "Development/".

src/templates/storyViewer.handlebars
cindyli: I agree this is a change for the better. But as the "back" button is only hidden at the first publishing step, it makes more sense to programmatically control the true/false value of "aria-hidden" to be in sync with the showing and hiding of the "back" button.

PR #19, opened 2019-03-18
SJRK-239: Remove editing capabilities from production site

BlueSlug:
**THIS UPDATE REMOVES EDITING CAPABILITIES, THEY WILL NEED TO BE RE-ENABLED AFTER MERGING THIS WORK INTO THE PRODUCTION BRANCH**

This pull request:
- adds a flag to disable the saveStoriesWithBinaries handler from doing its job. Instead the server returns an error message,
- hides the editor from the landing/edit page,
- hides the "publish my story" button from the editor, just in case,
- removes links to the Edit page, and
- updates the intro text so it no longer encourages sharing stories


BlueSlug:
@cindyli this is ready for review, but the pull request for CI hasn't yet been merged at this time:
https://github.com/inclusive-design/ci-service/pull/34

Perhaps we should coordinate with @waharnum to ensure everything goes as smoothly as possible

cindyli:
@BlueSlug, thanks for the implementation.

The change so far controls the server code with the `savingEnabled` flag but not the UI code. The UI side of enabling and disabling save is still hard-coded and requires modifications to a few files when the save needs to be re-enabled.

I think what we want is to have one flag that controls both server and UI sides. Turning the flag value is all required to do for enabling/disabling the save.

Some thoughts:
1. The `savingEnabled` config needs to be shared by both server and UI code.
2. Two set of text that are displayed on the landing page when save is on/off can all be saved in the database. The proper set will be retrieved based on the `savingEnabled` value.
3. The html tag attributes for hiding/showing divs/links should be dynamically injected using methods such as [a if check in the template](https://github.com/fluid-project/sjrk-story-telling/blob/stories-floe-dev/ui/src/templates/storyBlockMediaEdit.handlebars#L80-L89).

Let me know what you think. Thanks.

BlueSlug:
@cindyli thanks for your review! I agree that the flag should be accessible within the template, though there are a couple of issues that make this a little tricky to achieve.

The flag is currently kept in the `sjrk.storyTelling.server` grade, which means it is not accessible by the UI. I gave some thought to this when putting the flag where I did, keeping in mind what [SJRK-237](https://issues.fluidproject.org/browse/SJRK-237) means for global configuration of the site. I wasn't sure if I should make it a more globally-accessible value at this point, or if I did what the best approach would be given we plan to make a similar change in another Jira.

Additionally, I was trying to consider a way to conditionally merge in parts of a grade or entire grades. [ContextAwareness](https://docs.fluidproject.org/infusion/development/ContextAwareness.html) may be an option for this, though it will add some complexity to the project.

BlueSlug:
As discussed in person, I will attempt to do part of the work described in SJRK-237 involving generalizing the templates in order to get rid of the injected markup defined in `storyTellingServerUI.js`, and then adding a value into the Edit page grade which can be checked in the template (like @cindyli suggests above). Tests also

BlueSlug:
@cindyli ready for review once again. It's a bit more this time than usual.

Thanks!

ui/src/learningReflections/js/learningReflections.js
```diff
@@ -171,4 +171,15 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         }
     });
 
+    sjrk.storyTelling.learningReflections.storyEdit.hideEditorMarkup = function (savingEnabled, mainContainer, pageContainer) {
+        if (!savingEnabled) {
+            $(mainContainer).prop("hidden", true);
+            $(pageContainer).css({
```
cindyli: Please move the hardcoded css into a css class that can be applied when needed.

server/src/js/serverSetup.js
```diff
@@ -23,7 +23,8 @@ fluid.defaults("sjrk.storyTelling.server", {
                     binaryUploadDirectory: "./uploads",
                     uploadedFilesHandlerPath: "/uploads",
                     deletedFilesRecoveryPath: "/deleted_uploads",
-                    secrets: "@expand:sjrk.storyTelling.server.resolveJSONFile(./secrets.json)"
+                    secrets: "@expand:sjrk.storyTelling.server.resolveJSONFile(./secrets.json)",
+                    savingEnabled: false
```
cindyli, line 27: It would be great to have tests for savingEnabled === true/false.

BlueSlug:
@cindyli I've made the changes we discussed to the UI-side editor hiding code, and I've added tests to make sure everything on the Edit page has the correct visibility. The server-side tests with storage will fail (as in crash, not fail to pass) if the server-side flag is set to false. For this reason, I've set the test config to have saving enabled (I made that change in 3c03165b70608dd3b61ee933e5b72a32f02930b1). Should I try to find a way to test the server in both cases? This would require changing the way the server flag works, I think.

cindyli:
Would the server crashes when it's actually running (using `node index.js`) with "savingEnabled" set to false? If the server crashes at running certain test, the same issue might happen in real time when the same test condition is met.

ui/src/js/storyTelling-page-storyEdit.js
```diff
@@ -14,17 +14,8 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-                    "grid-template": "none"
-                },
-                show: {
-                    "padding-top": "1rem",
-                    "grid-template-columns": "2fr 3fr"
-                }
-            },
-            hideProperty: "hidden"
+            hiddenEditorClass: "hidden",
+            hiddenEditorProperty: "hidden"
```
cindyli: It's odd to optionalize a html element attribute. What other options can this property be? I think you can remove this line and directly use "hidden" attribute at line 212.

BlueSlug:
@cindyli the request works just fine with the flag set to either value in actual usage, since the function that handles the request checks the flag and stops there if it's not truthy:
https://github.com/BlueSlug/sjrk-story-telling/blob/SJRK-239/server/src/js/requestHandlers.js#L127

The tests, on the other hand, are checking for a particular response from that request. I wasn't able to determine a neat way to conditionally run those tests or conditionally alter the expected results, short of using context awareness.
see:
https://github.com/BlueSlug/sjrk-story-telling/blob/SJRK-239/tests/testServerWithStorage.js#L326
https://github.com/BlueSlug/sjrk-story-telling/blob/SJRK-239/tests/testServerWithStorage.js#L443-L472

I think from a semantic standpoint, it's a bit odd to run the storage tests if saving is turned off. I'll think about this some more.

BlueSlug:
Thinking about this further, I suppose I could use a similar solution to the requestHandler and check for the correct response values depending on the flag.

BlueSlug:
@cindyli I believe I have achieved what I described in my previous comment. I had to issue the same number of assertions, since that value is configured in a grade, so there are a couple of "empty" assertions.

tests/testServerWithStorage.js
```diff
@@ -317,11 +316,11 @@ sjrk.storyTelling.server.testServerWithStorageDefs = [{
     }, {
         event: "{storySave}.events.onComplete",
         listener: "sjrk.storyTelling.server.testServerWithStorageDefs.testStoryPostRequestSuccessful",
-        args: ["{arguments}.0", "{arguments}.1", "{that}.events.onStorySaveSuccessful"]
+        args: ["{arguments}.0", "{arguments}.1", "{that}.events.onStorySaveSuccessful", "{that}.configuration.server.options.globalConfig.savingEnabled"]
```
cindyli, line 319: The server test for the "savingEnabled" flag can be refactored better.

Right now, the "savingEnabled" value is dependent on what's set in the actual server source code "serverSetup.js", in which "savingEnabled" only has one value (true or false) at any moment. This means, running this test file only tests one flag value at any time rather than testing both cases. To fix it: 

1. No need to share the same set of test definition for 2 scenarios with a number of empty assertions. The current test definition is good for testing "true" scenario. Create a separate test definition for "false" scenario. In the new test, only responses from post story requests are checked to ensure an error is returned. The following checks for the saved story and the uploaded image are unnecessary.

2. An example of passing "savingEnabled" flag to the server component in test definitions:
```
sjrk.storyTelling.server.testServerWithStorageDefsWithSavingEnabled = [{
    name: "Test server with storage - savingEnabled is turned on",
    expect: 16,
    distributeOptions: {
        record: true,
        target: "{that server}.options.globalConfig.savingEnabled"
    }
...
}];
```
Please use a common variable to hold parts that can be shared by both definitions.

tests/ui/js/storyTelling-page-storyEditTests.js
```diff
@@ -14,7 +14,12 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     fluid.defaults("sjrk.storyTelling.page.testStoryEdit", {
         gradeNames: ["sjrk.storyTelling.page.storyEdit"],
         pageSetup: {
-            resourcePrefix: "../.."
+            resourcePrefix: "../..",
+            savingEnabled: false
```
cindyli, line 18: Again, this only tests the case when savingEnabled === false. The "true" case also needs to be tested. The same comment applies to other UI tests. Thanks.

ui/src/css/storyTelling.css
```diff
@@ -48,7 +48,21 @@ fieldset ul {
 
 .sjrk-st-page-content-container {
     background-color: #fff;
-    padding: 2rem;
+    padding: 0 2rem 2rem 2rem;
+}
+
+/* These sections are shown or hidden based on the theme's styling rules */
```
cindyli, line 54: Could you clarify this comment, regarding which themes are sharing this css flle and which theme will override these rules etc. Does "These sections" mean starting from here till the end of the file?

ui/src/css/storyTelling.css
```diff
@@ -48,7 +48,21 @@ fieldset ul {
...
+.sjrk-st-page-header-container,
```
cindyli, line 55: Is this rule to hide these containers created because these containers are not expected to show in Karisma theme? Can you add a comment to explain? Thanks.

cindyli:
I noticed when the saving/editing is disabled, once users leave the home page "storyEdit.html", they can never go back to it because the "build your story" button is now invisible.

When the editing is disabled, this "story edit" page now serves as a home page rather than story editing. It's weird to not be able to access it once leaving for other pages. 

One possibility is to allow clicks on SJRK logo at the header to lead back to the home page.

This needs to be discussed with designers. I don't expect it to be solved in this pull request. Thanks.

BlueSlug:
As discussed with @cindyli and @michelled, the outstanding review items have been captured in a new Jira issue to be addressed at a later time: https://issues.fluidproject.org/browse/SJRK-241

PR #22, opened 2019-04-10
SJRK-227: Added logging invoker and listeners for all datasource.onError

BlueSlug:
This pull request does not include integration tests as I was unable to determine how to listen for the fluid.loggingEvent being fired once fluid.log was called from the test server.

### DIFF
--- a/server/src/js/dataSource.js
+++ b/server/src/js/dataSource.js
@@ -20,6 +20,16 @@ fluid.defaults("sjrk.storyTelling.server.dataSource.couch.base", {
         getURL: {
             funcName: "sjrk.storyTelling.server.dataSource.couch.base.getURL",
             args: ["{that}.options.host", "{that}.options.path"]
+        },
+        logDataSourceError: {
+            funcName: "fluid.log",
+            args: [fluid.logLevel.WARN, "Datasource error: ", "{arguments}.0"]
+        }
+    },
+    listeners: {
+        "onError.logDataSourceError": {
+            func: "{that}.logDataSourceError",
+            args: ["{arguments}.0"]
         }
     }
 });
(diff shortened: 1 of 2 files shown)


cindyli:
Why not directly add the error handler to the base data source component `sjrk.storyTelling.server.dataSource.couch.base`?

BlueSlug:
@cindyli a great suggestion :) I've implemented it and everything seems to be working the same. The error messages are slightly less detailed, now, but the key info is there (in this case, the webserver is unable to connect to the couch server)

server/src/js/dataSource.js
```diff
@@ -20,6 +20,16 @@ fluid.defaults("sjrk.storyTelling.server.dataSource.couch.base", {
...
             args: ["{that}.options.host", "{that}.options.path"]
+        },
+        logDataSourceError: {
+            funcName: "fluid.log",
+            args: [fluid.logLevel.WARN, "Datasource error: ", "{arguments}.0"]
+        }
+    },
+    listeners: {
+        "onError.logDataSourceError": {
+            func: "{that}.logDataSourceError",
```
cindyli: You can get rid of the `logDataSourceError` invoker by replacing line 31, 32 with line 25, 26.

BlueSlug: I agree that this will make the code shorter, but I'm unsure since that would make this particular task more closely coupled to the onError event itself. Then again, it is "just" a call to fluid.log. I can't think of a case at the moment where we'd want to call the invoker in a way that isn't associated with the onError event on this component.

cindyli:
Merged at 9448cb4f4372ca1f5afb79503ceb24b155d07e77

PR #23, opened 2019-05-06
SJRK-229: disabling the saving of empty blocks to fix graphical bug

BlueSlug:
The tests have been simplified using sequence grades, and though I'm sure there's more that could be done, I believe it would be better to move onto other issues.


tests/ui/js/storyTelling-page-storyEditTests.js
```diff
@@ -449,6 +884,10 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         jqUnit.assertEquals("The response text is as expected", expectedText, actualText);
     };
 
+    sjrk.storyTelling.page.storyEditTester.setCurrentBlock = function (testCaseHolder, currentBlock) {
+        testCaseHolder.options.members.currentBlock = currentBlock;
```
cindyli: Member options can be accessed as a direct link to the component, such as: `testCaseHolder.currentBlock = currentBlock;`.

Similarly, all appearances of `{storyEditTester}.options.members.currentBlock` at line 129, 151, 170 can be changed to  `{storyEditTester}.currentBlock`.

Also, this function can also be replaced by declarative statements, for example, line 114-116 can be changed to:
```
{
    "event": "{storyEdit}.storyEditor.events.onNewBlockTemplateRendered",
    listener: "fluid.set",
    args: ["{storyEditTester}", ["currentBlock"], "{arguments}.0"]
}
```

The same comment applies to other places where `setCurrentBlock()` is used.

tests/ui/js/testUtils.js
```diff
@@ -114,10 +114,20 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         return element.selectorName || element.selector || element.toString();
     };
 
-    sjrk.storyTelling.testUtils.checkFirstBlockCheckbox = function (blockManager) {
+    sjrk.storyTelling.testUtils.checkBlockCheckboxes = function (blockManager, options) {
```
cindyli: It seems the only option that `options` object accepts is `checkFirstBlock`. Would it be more straightforward and efficient to pass in `checkFirstBlock` as the 2nd argument rather than passing in an object?

ui/src/js/storyTelling-page-storyEdit.js
```diff
@@ -68,6 +68,34 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            making these values a configuration option in each block grade,
+            maybe even making the block contentString model relays based on it
+        */
+        requiredBlockValues: {
+            "text": ["heading", "text", "simplifiedText"],
+            "image": ["imageUrl"],
+            "audio": ["mediaUrl"],
+            "video": ["mediaUrl"]
+        },
+        modelRelay: {
```
cindyli, line 82: It might be possible to combine these 2 model relays into one by:
1. Call `sjrk.storyTelling.page.storyEdit.removeEmptyBlocks` in the "singleTransform" of "editorStoryToPreviewer" relay;
2. Get rid of "excludeEmptyBlocks" relay.

ui/src/js/storyTelling-page-storyEdit.js
```diff
@@ -207,6 +230,40 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+     * those values is truthy, the block is not empty.
+     * - "block": a story block
+     * - "requiredValues": an array of the "required" values of a block, according
+     *                     to the definition above
+     */
+    sjrk.storyTelling.page.storyEdit.isEmptyBlock = function (block, requiredValues) {
+        var isEmptyBlock = true; // assume the block is empty
+
+        fluid.each(requiredValues, function (requiredValue) {
+            if (block[requiredValue]) {
```
cindyli: The `fluid.each` iteration can be stopped once the first non-empty value is hit. What do you think to change line 260 to `return false`?

tests/ui/js/testUtils.js
BlueSlug: I think this is a case of adding a more flexible structure ahead of time for something I can't foresee. I don't need it for now, if ever, so I'll remove it :)

ui/src/js/storyTelling-page-storyEdit.js
BlueSlug: If i change line 260 to `return false`, I'm only returning the callback to fluid.each, aren't I? I think I had tried this when I first wrote these lines

cindyli: Good point and it's true. We won't be able to stop the iteration.

cindyli: Probably try fluid.find() to stop at the first non-empty value.

BlueSlug: I've made this change using fluid.find_if()

PR #30, opened 2019-06-10
SJRK-80: Icons and logos do not adapt to UIO high contrast

BlueSlug:
This pull request adds versions of each SVG to accommodate the existing UIO contrast themes, including newer ones that aren't present in the storytelling tool yet. It also converts several images used on the site from proper `img` elements to background images via CSS, enabling them to be swapped around on the fly. Several alternatives were explored, including putting the SVG elements directly in the template markup or using SVG/CSS filters to achieve the correct look, but the current approach is the most complete I was able to implement thus far.

most of the SVG files have also been simplified in order to take up less space and be more easily recoloured.

@cindyli to review.


ui/src/learningReflections/templates/learningReflections-storyEditor.handlebars
```diff
@@ -44,10 +44,10 @@
                     </div>
                 </div>
                 <div class="sjrk-st-button-container">
-                    <a href="javascript:void(0)" class="sjrk-st-story-editor-next sjrkc-st-story-editor-next">
+                    <button type="button" class="sjrk-st-story-editor-next sjrkc-st-story-editor-next" aria-role="{{localizedMessages.message_continue}}">
```
cindyli: It seems making more sense to use the keyword `aria-label` here for indicating the button usage rather than `aria-role`.

The same comment applies to all uses of `aria-role` below.

PR #33, opened 2019-07-17
SJRK-237: Reorganize project folder structure and theming

BlueSlug:
Substantial changes were made, please see [the Jira](https://issues.fluidproject.org/browse/SJRK-237) for more information.

Some tests were not added due to a race condition relating to a bug with UIO implementation. This has been captured in [SJRK-255](https://issues.fluidproject.org/browse/SJRK-255)

@cindyli to provide review :)


README.md
```diff
@@ -15,96 +15,48 @@ This project also uses [Node.js](https://nodejs.org), [Kettle](https://github.co
...
-- Add any new associated handlebars templates to be used by new UI components
-- If your new content contains any new wording, you can create new message bundle files or add them to the existing message files (in the `ui/src/messages` directory) and refer to those message names in the handlebars template(s). Please note that a `ui` component specifies and loads only one message bundle at a time.
+The "base" files included in this repository, located in the `themes/base` directory, are intended to provide a bare-bones implementation of the Storytelling Tool. The interface is extensible and customizable to enable the creation of new experiences, themes and story contexts with the tool. To create a custom theme, follow these steps:
+- Set the `theme` value in the server config file (see [Running the site](#Running-the-site)) to the name of your new theme. For the sake of this example, we'll use `cuteCats`.
+- Create a new folder for code and assets associated with the new theme (for examples, please see the `themes/karisma` or `themes/learningReflections` folders)
```
cindyli: Please also describe the name of the new folder needs to be same as the `theme` value defined in the server config file.

src/server/serverSetup.js
```diff
@@ -158,6 +180,18 @@ fluid.defaults("sjrk.storyTelling.server.app.storyTellingHandlers", {
             "route": "/admin/deleteStory/:id",
             "method": "get"
         },
+        testDataHandler: {
```
cindyli, line 173: What is `testDataHandler` for? There isn't a /testData sub-directory in the sjrk root directory.

src/server/serverSetup.js
```diff
@@ -114,18 +119,35 @@ fluid.defaults("sjrk.storyTelling.server", {
...
                         options: {
-                            "root": "./ui",
+                            "root": "./src/ui"
+                        }
+                    },
+                    // the custom theme for the site, loaded in "on top" of base
+                    currentTheme: {
+                        type: "kettle.middleware.static",
+                        options: {
+                            root: "@expand:sjrk.storyTelling.server.getCustomThemeFolder({server}.options.globalConfig.theme, ./themes/%theme)",
```
cindyli: The base theme directory `./themes` can be defined as a configurable component option.

src/server/serverSetup.js
```diff
@@ -196,3 +229,19 @@ sjrk.storyTelling.server.resolveJSONFile = function (jsonFilePath) {
...
+sjrk.storyTelling.server.getCustomThemeFolder = function (theme, themeFolder) {
+    var folder = ".";
+
+    if (theme) {
+        var resolvedFolder = fluid.stringTemplate(themeFolder, { theme: theme });
+        // ensure the folder exists, otherwise return an error
+        if (fs.existsSync(resolvedFolder)) {
+            folder = resolvedFolder;
+        } else {
+            folder = undefined;
```
cindyli: When the custom theme folder is not found, starting the server will throw a fatal error. It would be more user friendly to report a proper error message using fluid.fail() and quit.

src/ui/storyTellingServerUI.js
```diff
@@ -0,0 +1,143 @@
...
+                "record": retrievedStory
+            };
+
+            var storyViewComponent = sjrk.storyTelling[theme].storyView(options);
+            storyPromise.resolve(storyViewComponent);
+        }).fail(function (error) {
+            storyPromise.reject(error);
+        });;
+    } else {
+        storyPromise.reject();
```
cindyli: Please provide a proper error message. Also, can we use a common structure for all returned errors:
```
{
    isError: true,
    message: {the-actual-error-message}
}
```
to distinguish errors and non-errors. This structure has already been used in [staticMiddlewareSubdirectoryFilter.js](https://github.com/BlueSlug/sjrk-story-telling/blob/SJRK-237/src/server/middleware/staticMiddlewareSubdirectoryFilter.js#L40).

cindyli: This comment applies to the entire code base. Thanks.

src/ui/storyTellingServerUI.js
```diff
@@ -0,0 +1,143 @@
...
+ *               regardless of whether theme information was specified or retrieved
+ * - "themeOverride": allows overriding of the theme stored in the configuration
+ */
+sjrk.storyTelling.loadThemedPage = function (callback, themeOverride) {
+    var loadPromise = fluid.promise();
+
+    var callbackFunction = typeof callback === "function" ? callback : fluid.getGlobalValue(callback);
+
+    $.get("/clientConfig").then(function (data) {
+        var theme = themeOverride ? themeOverride : data.clientConfig ? data.clientConfig.theme : "base";
```
cindyli: The fall-back logic to the base theme should be on the server side as the first thing when global config is read in, which can then be used by the following code as a proper theme. It's better to handle this kind of case as early as possible.

src/ui/storyTellingServerUI.js
```diff
@@ -0,0 +1,143 @@
...
+
+        if (theme && theme !== "base") {
+            return sjrk.storyTelling.loadCustomThemeFiles(callbackFunction, theme).then(function () {
+                loadPromise.resolve(theme);
+            }, function (error) {
+                loadPromise.reject(error);
+            });
+        } else {
+            callbackFunction("page");
```
cindyli: Why is the parameter "page" for loading the base theme?

README.md
```diff
@@ -15,96 +15,48 @@ This project also uses [Node.js](https://nodejs.org), [Kettle](https://github.co
...
+* Run a CouchDB server on `localhost:5984`
+    * Docker is an easy way to do this, we recommend the `apache/couchdb` image. To use that image via Docker, run this command: `docker run -p 5984:5984 -d apache/couchdb`
+* Run `node .\src\server\db\dbSetup.js` to configure necessary CouchDB databases (this will also ensure your CouchDB instance is set up in [single-node mode](https://docs.couchdb.org/en/stable/setup/single-node.html))
     * If you have an [admin user](https://docs.couchdb.org/en/stable/intro/security.html) configured on you CouchDB instance, you will need to provide the credentials to `dbSetup.js`. This can be done by setting the `COUCHDB_URL` environment variable. `dbSetup.js` will use the URL specified in `COUCHDB_URL` when connecting to CouchDB. For example: `COUCHDB_URL=http://admin:adminpassword@localhost:5984`
+* Create a `sjrk.storyTelling.server.config.json5` file in the style of `sjrk.storyTelling.server.config.json5.example`. This file configures the server and allows you to customize the current theme or configure a new one. Edit the values in the `globalConfig` section to do this. More on this in [Theme Customization](#Theme-Customization).
```
cindyli: It would be helpful to create a section to describe all properties and their purpose, that can specified in `globalConfig`.

src/ui/storyTellingServerUI.js
```diff
@@ -0,0 +1,143 @@
...
+/* Loads CSS and JavaScript files for the provided theme into the page markup.
+ * If JavaScript file loading is successful, the callback function is called.
+ * - "theme": the theme of the story Browse page (pass "page" in for base theme)
+ * - "callback": a function to call once everything has completed
+ */
+sjrk.storyTelling.loadCustomThemeFiles = function (callback, theme) {
+    var loadPromise = fluid.promise();
+
+    var cssUrl = fluid.stringTemplate("/css/%theme.css", {theme: theme});
+    var scriptUrl = fluid.stringTemplate("/js/%theme.js", {theme: theme});
```
cindyli: It seems limited to only allow users to write one js file and one css file for their custom themes. Users would have more freedom if these type of files can be included in their custom templates such as a header template.

Since this approach satisfies the requirement for all two custom themes we have so far, let's stay with allowing only one for each. It can be improved when the requirement goes beyond its limit.

tests/server/testServerWithBaseTheme.js
```diff
@@ -0,0 +1,109 @@
...
+            }
+        }
+    },
+    sequence: [{
+        // test the clientConfig info endpoint
+        func: "{clientConfigRequest}.send"
+    }, {
+        event: "{clientConfigRequest}.events.onComplete",
+        listener: "sjrk.storyTelling.server.verifyGetClientConfigSuccessful",
+        args: ["{arguments}.0", "{arguments}.1", undefined]
```
cindyli: As said earlier, the fall back to the base theme should have already been kicked in here.

src/ui/storyTellingServerUI.js
cindyli: I see. Using the parameter "page" is for loading the default component used by the base theme such as "sjrk.storyTelling.page.storyBrowse". Can you add a comment to explain? Thanks.

tests/server/testServerWithBaseTheme.js
```diff
@@ -0,0 +1,109 @@
...
+        // test getting a file that exists only for custom theme, with no custom theme specified
+        func: "{baseThemeFileMissingRequest}.send"
+    }, {
+        event: "{baseThemeFileMissingRequest}.events.onComplete",
+        listener: "sjrk.storyTelling.server.verifyGetThemeFileUnsuccessful",
+        args: ["{arguments}.0", "{arguments}.1"]
+    }]
+}];
+
+kettle.test.bootstrapServer(sjrk.storyTelling.server.testServerWithBaseThemeDefs);
```
cindyli, line 109: We should test rendering of actual theme UIs. Is it where you have issues with UIO - https://issues.fluidproject.org/browse/SJRK-255?

src/server/serverSetup.js
BlueSlug: There's a testData directory in the /tests/ directory, which contains some files used in both the node and browser tests. The contents are served so the browser tests can refer to them. I didn't want to serve/expose the entire /tests/ directory, but considering we're keeping the node and browser tests separate, I didn't want to keep the testData directory in one or the other of those two locations.

src/server/serverSetup.js
BlueSlug: Do you mean to suggest that I could configure the root directory for the baseTheme middleware? In that case I intentionally chose not to do this, in order to avoid confusion on the part of someone else who would like to implement their own custom theme. I could make it a configurable option and simply not pull that from `sjrk.storyTelling.server.config.json5`, but I thought this was a reasonable implementation, since I only wanted to change it for node tests.

Alternatively, are you suggesting that in this particular case that the second argument for `getCustomThemeFolder()` should be an option instead of hardcoded here?

tests/server/testServerWithBaseTheme.js
BlueSlug: Yes, the tests I was trying to add when I ran into the SJRK-255 issue were to add calls to `sjrk.storyTelling.loadStoryFromParameter()` and `sjrk.storyTelling.loadBrowse()`, which would each load in actual themed UIs.

Mysteriously, while working on SJRK-255 today, I found myself unable to reproduce the error.

src/ui/storyTellingServerUI.js
BlueSlug: I agree, and I've filed [SJRK-256](https://issues.fluidproject.org/browse/SJRK-256) to keep track of this suggestion. :)

src/server/serverSetup.js
cindyli: Sorry about the confusion. The "configurable" I meant was to define the base theme dir "./themes" as a component option rather than via `sjrk.storyTelling.server.config.json5` for end users. I think this flexibility will make it easier for reusing or configuring the component as well as writing tests as you said.

src/server/serverSetup.js
cindyli: Got it. Thanks.

README.md
```diff
@@ -5,106 +5,66 @@ https://sojustrepairit.org/
...
+The tool is split up into two main pieces: a webserver to manage communication with the CouchDB database, and an HTML interface for creating and sharing stories.
+* Server code can be found in `src/server`
+* Common JavaScript code can be found in `src/ui`, though a significant portion of the code may be present in the theme folders (see [Theme Customization](#Theme-Customization)).
+
+## Configuring the application
+While Infusion allows for just about any possible extension you can imagine and implement, there are a few handy configuration settings provided as part of the code which allow you, for instance, to customize the current theme or configure a new one. The server will load a file named `sjrk.storyTelling.server.config.json5`, and the `globalConfig` collection is where you will find the following settings:
+
+| Setting | Description |
+|---------|-------------|
+| `port`  | _(Optional)_ The HTTP port the server will be hosted on. The default value is 8081. |
```
cindyli: Looking at the source code, `globalConfig` also supports: binaryUploadDirectory, uploadedFilesHandlerPath, deletedFilesRecoveryPath, secrets.adminPass.

Since `sjrk.storyTelling.server.config.json5` is in json5 format, it might be helpful to write a comment section at the start to list all supported values and their goals, as a reminder to end users and developers.

README.md
```diff
@@ -5,106 +5,66 @@ https://sojustrepairit.org/
...
+| `savingEnabled` | Specifies whether editing and saving are allowed. If the value is set to `false`, then the site is effectively in "read-only" mode and will behave as a collection of stories rather than an authoring tool. |
+| `theme` | _(Optional)_ The theme to load the site with. If this isn't specified, a base theme will be loaded. More on this in [Theme Customization](#Theme-Customization). |
```
cindyli: Please also talk about all supported themes: base, karisma, learningReflections.

README.md
```diff
@@ -5,106 +5,66 @@ https://sojustrepairit.org/
...
+| `themeIndexFile` | _(Optional)_ The file to serve at the site root. E.g. `"themeIndexFile": "index2.html"`. The default is `storyBrowse.html` |
```
cindyli: The statement of "The default is `storyBrowse.html`" seems inaccurate. My understanding by looking at the code is:
1. when the "base" theme is used, the value of "themeIndexFile" will be ignored and be automatically set to storyBrowse.html
2. when a custom theme is used, the value of "themeIndexFile" must be provided.

Please correct me if I'm wrong. If there are such details, better to describe them to avoid confusion. Thanks.

cindyli:
UIO is broken with all three themes. The console error shows `TypeError: text is undefined`.

Screen Shot 2019-07-19 at 10 42 44 AM

README.md
BlueSlug: The three upload-related keys (`binaryUploadDirectory`, `uploadedFilesHandlerPath`, and `deletedFilesRecoveryPath`) were previously pulled in from the configs file, but after further thought I decided these are not "quick and easy" things to configure. If these values are changed, then the code needs to be updated to create and include the referenced directories, as well as to ensure they have the correct permissions set. This means it's non-trivial, and so I think if anyone is going to be digging that far into customizing the application, they'll be able to find the values in `serverSetup.js`. Is that a reasonable assumption?
I suppose I could put comments in the JS to explain their purpose, if that helps.

As for `secrets.adminPass`, this is something I believe we've discussed in the past. The admin password is something that should be kept closely guarded, and having it alongside more "open" values may be misleading or lead to it being overlooked. I suppose I could keep it at the `secrets` path within `globalConfig` to make that more obvious, but I think we decided that having it in a separate file means that the developer will make a conscious move to create that file and keep track of it separately. It's not a perfect solution in either case, but it may prevent accidental disclosure. I'm not 100% sold on this idea haha

README.md
BlueSlug: The case you've described in `1.` is accurate; if `theme` isn't specified, it will load `storyBrowse.html` regardless of whether `themeIndexFile` is specified. I wonder if it makes sense to have `themeIndexFile` control the root file that gets served in all cases.
As for `2.`, the site will work if `theme` is specified and `themeIndexFile` is not. In this case, it will load the `baseTheme` index file (which is `storyBrowse.html`) but with the custom theme applied.

BlueSlug: Would it be better to add a note describing how `themeIndexFile` will be ignored if `theme` is not present and file a Jira for the improvement, or should I try to make it apply regardless of whether a custom theme is loaded?

README.md
cindyli: If these properties are not intended to be configured by end users, they probably should be moved out of `globalConfig` into component options or even member options to prevent them from being accidently overwritten or potential security issues with the present of `adminPass`.

README.md
cindyli: Thanks for the clarification. Yes, it would be better to describe how `themeIndexFile` works as well as the fact that the "base" theme is the fall back theme when `theme` is not specified.

Can you elaborate on "should I try to make it apply regardless of whether a custom theme is loaded"? 

Thanks.

BlueSlug: Sure :) what I meant by "apply regardless of whether a custom theme is loaded" is to implement the `themeIndexFile` (maybe renaming it to simply `indexFile` or `rootFile` or `defaultDocument`) in such a way that it specifies the file to serve at the site root regardless of what theme is used. I.e. if the developer wants to use `base` but change the default file from `storyBrowse.html` to, say, `storyEdit.html`, they can use the same key in `globalConfig` instead of altering `serverSetup.js`

BlueSlug:
> UIO is broken with all three themes. The console error shows `TypeError: text is undefined`.

I'm having trouble reproducing this issue, and though UIO is not loading for me either, the error message I'm seeing is consistent with the one I reported in SJRK-255:
```
fluidParser.js:420 Uncaught TypeError: Cannot read property 'replace' of undefined
    at Object.fluid.XMLEncode (fluidParser.js:420)
    at renderComponent (fluidRenderer.js:980)
    at renderComponentSystem (fluidRenderer.js:1139)
    at renderRecurse (fluidRenderer.js:1327)
    at Object.that.renderTemplates (fluidRenderer.js:1405)
    at Object.fluid.reRender (fluidRenderer.js:1488)
    at Object.fluid.render (fluidRenderer.js:1555)
    at Object.that.render (RendererUtilities.js:104)
    at fluid.rendererComponent.refreshView (RendererUtilities.js:178)
    at invokeInvoker (FluidIoC.js:1749)
```

cindyli:
I encountered the same error you had when using Chrome. The error `TypeError: text is undefined` is with Firefox.

BlueSlug:
I was just testing in Firefox when you replied :) Okay, well in either case it seems to be related to UIO attempting to render before its localized messages have been fully loaded, which I intended to tackle in SJRK-255. How should I proceed?

cindyli:
I feel we should tackle it as part of this reorganization work. It won't be pretty if the live SJRK site has a broken UIO.

README.md
cindyli: This change doesn't seem necessary at least for now. Let's stick with `themeIndexFile` and have a clear document for it. Thanks.

Gruntfile.js
```diff
@@ -16,10 +16,10 @@ module.exports = function (grunt) {
         // Project package file destination.
         pkg: grunt.file.readJSON("package.json"),
         eslint: {
-            all: ["server/src/**/*.js", "ui/src/**/*.js", "tests/**/*.js", "*.js"]
+            all: ["src/**/*.js", "tests/**/*.js", "*.js"]
         },
         jsonlint: {
-            all: ["package.json", ".eslintrc.json", "server/src/**/*.json", "ui/src/**/*.json", "tests/**/*.json"]
+            all: ["package.json", ".eslintrc.json", "themes/**/*.json", "tests/**/*.json"]
```
cindyli: Json5 files such as "sjrk.storyTelling.server.config.json5" and those in the "tests/" directory should be linted too.

README.md
```diff
@@ -5,106 +5,73 @@ https://sojustrepairit.org/
...
+- Create a new folder for code and assets associated with the new theme (for examples, please see the `themes/karisma` or `themes/learningReflections` folders). The folder name **must** be the same as the `theme` value in the config file.
+- Create a JavaScript file which contains extensions of the `page` grades (e.g. for `sjrk.storyTelling.page.storyEdit`, add a new grade `sjrk.storyTelling.cuteCats.storyEdit` which has the former as one of its gradeNames). These extensions could include new `ui` components for new sections of the page, new events or other functionality. The name of the JavaScript file must match the folder and theme name, e.g. `themes/cuteCats/js/cuteCats.js`
+- Create a CSS file in the `css` directory with all of the styling rules specific to the new theme. The CSS file, like the JavaScript file, should have the same name as the theme and theme folder: `themes/cuteCats/css/cuteCats.css`
+- Add any new associated [handlebars](https://handlebarsjs.com/) templates to be used by new UI components
```
cindyli: Indicate these templates need to be in the `templates` directory.

sjrk.storyTelling.server.config.json5.example
```diff
@@ -0,0 +1,25 @@
+{
+    "type": "sjrk.storyTelling.server.config",
+    "options": {
+        "gradeNames": ["sjrk.storyTelling.server"],
+        "components": {
+            "server": {
+                "options": {
+                    "globalConfig": {
+                        // the HTTP port the server will hosted from, default is 8081
```
cindyli: will -> will be

sjrk.storyTelling.server.config.json5.example
```diff
@@ -0,0 +1,25 @@
...
+                        "port": "8081",
+
+                        // the custom theme to load in on top of the base theme
+                        "theme": "learningReflections",
+                        
+                        // a default file to load at the site root (e.g. index.html)
+                        "themeIndexFile": "introduction.html",
+
+                        // if true, saving stories to the database will be enabled
```
cindyli, line 18: As you said in the readme, this flag is actually to enable/disable the authoring part. It's probably more accurate to call it "authoringEnabled".

BlueSlug:
Previously identified, but a good time to tackle it: serverSetup.js dataSource dist opts duplication - make it more DRY by either adding it to the base grade or the parent component

src/server/serverSetup.js
```diff
@@ -168,31 +205,53 @@ fluid.defaults("sjrk.storyTelling.server.app.storyTellingHandlers", {
...
+ * - "themeFolder": The folder/path that contains the theme being retrieved
+ */
+sjrk.storyTelling.server.getThemePath = function (theme, themeFolder) {
+    var themePath = ".";
+
+    if (theme) {
+        themePath = fluid.stringTemplate(themeFolder, { theme: theme });
+
+        if (!fs.existsSync(themePath)) {
+            fluid.fail("The custom theme folder " + themePath + " does not exist. Please verify that the theme name is configured properly.");
```
cindyli, line 239: The failure reported will include your message as well as the detailed tech error by printing out the component tree. The way I found to report a graceful error and exit is to replace this line with:

```
console.log("FATAL ERROR: ...");
process.exit(1);
```

src/ui/storyTellingServerUI.js
```diff
@@ -0,0 +1,175 @@
...
+            return sjrk.storyTelling.loadCustomThemeFiles(callbackFunction, theme).then(function () {
+                loadPromise.resolve(theme);
+            }, function (jqXHR, textStatus, errorThrown) {
+                loadPromise.reject({
+                    isError: true,
+                    message: errorThrown
+                });
+            });
+        } else {
+            theme = "base";
```
cindyli: This line is unnecessary. All exceptions such as undefined theme should have been handled at the server side. This "if...else..." block can simply check `theme === "base"` or else.

src/ui/storyTellingServerUI.js
```diff
@@ -0,0 +1,175 @@
...
+            callbackFunction(theme);
+            loadPromise.resolve(theme);
+        }
+    }, function (jqXHR, textStatus, errorThrown) {
+        callbackFunction("base"); //default to the base theme page
```
cindyli: replace "base" with the variable `theme` in case there's a name change for the base theme in the future. Please avoid hardcoded strings whenever possible. Thanks.

tests/server/configs/sjrk.storyTelling.server.testServerWithBaseTheme.json5
```diff
@@ -0,0 +1,29 @@
+{
+    "type": "sjrk.storyTelling.server.testServerWithCustomTheme",
+    "options": {
+        "gradeNames": ["sjrk.storyTelling.server"],
+        "components": {
+            "server": {
+                "options": {
+                    "globalConfig": {
+                        "binaryUploadDirectory": "./tests/server/uploads",
```
cindyli: Apparently tests don't make use of config values at line 9-10. Otherwise they won't pass with these configs being moved to `secureConfig`.

It would be nice to have integration tests to cover the communication between server and client so these configs make sense in tests.

tests/server/configs/sjrk.storyTelling.server.testServerWithBaseTheme.json5
```diff
@@ -0,0 +1,29 @@
...
+                        "uploadedFilesHandlerPath": "./tests/server/uploads"
+                    },
+                    "components": {
+                        "storyDataSource": {
```
cindyli: This subcomponent doesn't seem to be used in the test too. It would be less confusing to only define the necessary bits that have been used by the test.

This comment applies to other config files too. Thanks.

tests/ui/js/storyTellingServerUITests.js
```diff
@@ -0,0 +1,173 @@
...
+            if (index === 4) {
+                sjrk.storyTelling.testUtils.setQueryString("testParameterFromUrl=testValue&emptyTestParameterFromUrl=");
+            }
+
+            var actualResult = sjrk.storyTelling.getParameterByName(testCase.parameter, testCase.url);
+            jqUnit.assertEquals("Query string parameter '" + testCase.parameter + "' is retrieved as expected", testCase.expected, actualResult);
+        });
+    });
+
+    // var browsePageOptions = {
```
cindyli: Are these commented lines are for the test meant to be dealt with at https://issues.fluidproject.org/browse/SJRK-255?

src/ui/storyTellingServerUI.js
```diff
@@ -0,0 +1,175 @@
...
+
+/* Loads custom theme files if and only if one of either the theme override or
+ * server configuration theme are present AND not set to the base theme.
+ * Once complete, the provided callback function is called and the theme name is
+ * passed into it. If custom theme files are not loaded, the "base" theme is passed.
+ * - "callback": a function to call once everything has completed. Will be called
+ *               regardless of whether theme information was specified or retrieved
+ * - "themeOverride": allows overriding of the theme stored in the configuration
+ */
+sjrk.storyTelling.loadThemedPage = function (callback, themeOverride) {
```
cindyli: It's weird to accept "themeOverride" parameter that could override the theme sent back from the server, which is the theme defined in the config file. As what's shown in your test case, what would happen if this value is provided as `null` or `undefined`? Does the client side have to perform its own fall back to the base? 

Could you elaborate the purpose of "themeOverride" parameter? Thanks.

Gruntfile.js
BlueSlug: jsonlint does not currently support json5 format, nor am I able able to find a viable eslint plugin for json5. That said, if I remove the comments from the json5 files, they should pass jsonlint, but if any developer decides to use any of the extra features of json5 they will fail the linting checks.

sjrk.storyTelling.server.config.json5.example
BlueSlug: I think the comment in the example json5 file is actually more accurate, and the comment in the readme is descriptive of the behaviour of _both_ the server flag and the UI-side flag, which are still separately defined. I will add a comment in the readme to indicate that both should be set together and should match each other.

Gruntfile.js
cindyli: The universal repo has [an example](https://github.com/GPII/universal/blob/master/Gruntfile.js#L23) of linting json5 files.

sjrk.storyTelling.server.config.json5.example
cindyli: Why does UI side need its own flag? /clientConfig is returning "savingsEnabled" value that is read from the server side config file.

Gruntfile.js
amb26: The relevant module is https://github.com/GPII/gpii-grunt-lint-all , written by @the-t-in-rtf which is well worth using

src/ui/storyTellingServerUI.js
BlueSlug: In this case the request to `/clientConfig` on the server will have failed, so `theme` will be undefined (it's out of scope in either case). Would it be acceptable to pass in something like `baseThemeName`  in the call to `loadThemedPage()` and use that here? Likewise, I could make that an option in the `secureConfig` collection and refer to that whenever it needs to be used in order to avoid hardcoded strings.

tests/ui/js/storyTellingServerUITests.js
BlueSlug: That's correct, and the same is true of the last two elements of the test sequence

src/ui/storyTellingServerUI.js
BlueSlug: I had added that in order to do more robust testing of the `loadThemedPage` function, since it's a browser test and not based on varying server config files. This is the only usage of it that parameter that I've anticipated. If `themeOverride` is falsy, it will fall back to the theme returned from the request to `/clientConfig`, according to line 118.

I suppose the structure of this function should be reworked, if `themeOverride` is going to be kept, since some of the possible cases are not handled, namely if the request fails but `themeOverride` is truthy, or if the request fails and `themeOverride` is truthy but also not valid.

src/server/serverSetup.js
BlueSlug: I implemented this suggestion, and while it did kill the server gracefully, it was not something I was able to test against to ensure that this was what was happening. Since the `process.exit(1)` call simply kills the server, I would have to use or write some sort of external tool to test the output in the console.

The detailed component tree printout is verbose and would make finding the actual error message difficult, but I think there's value in doing things the "Infusion way" and calling fluid.fail. I wonder if there is a way to call fluid.fail without having it print out the component tree, but just the message being passed into it.

@amb26 do you have any thoughts on this matter?

BlueSlug:
> Previously identified, but a good time to tackle it: serverSetup.js dataSource dist opts duplication - make it more DRY by either adding it to the base grade or the parent component

I was not able to achieve this. In all cases, the server would fail to load.

src/ui/storyTellingServerUI.js
cindyli: If the server doesn't work, would the client be able to operate on its own? If not, it might be more helpful to report an error rather than carrying on the operation.

src/ui/storyTellingServerUI.js
cindyli: A better way to test it is to create a mocking server response for /clientConfig, which mocks http responses you are expecting for your test cases. Infusion uses "Sinon.js" library that also has [the feature of creating a fake server](https://sinonjs.org/#fake-server).

cindyli:
> > Previously identified, but a good time to tackle it: serverSetup.js dataSource dist opts duplication - make it more DRY by either adding it to the base grade or the parent component
> 
> I was not able to achieve this. In all cases, the server would fail to load.

I issued [this pull request](https://github.com/BlueSlug/sjrk-story-telling/pull/25) to your SJRK-237 branch to address the issue.

BlueSlug:
> > > Previously identified, but a good time to tackle it: serverSetup.js dataSource dist opts duplication - make it more DRY by either adding it to the base grade or the parent component
> > 
> > 
> > I was not able to achieve this. In all cases, the server would fail to load.
> 
> I issued [this pull request](https://github.com/BlueSlug/sjrk-story-telling/pull/25) to your SJRK-237 branch to address the issue.

That was the one case I had neglected to try :0 I'd tried things that mostly involved incorporating the expander into the base grade and not the parent component.

Thanks @cindyli!

BlueSlug:
> What do you think about calling it "base" rather than "baseTheme"?

I thought about this a little bit before arriving at `baseTheme`, and the main reason is that we have other parts of the codebase (heh) where we use the term "base" to indicate a grade that is used primarily as the common source of multiple other grades. While the purpose of the base theme is similar, it also stands alone, so I thought it would be better to be explicit in naming it after a theme. This also helps indicate where the theme name would go for any new themes (this is reflected in the readme)

BlueSlug:
@cindyli ready for further review. In my testing I discovered an issue with the empty block filter and captured it as [SJRK-263](https://issues.fluidproject.org/browse/SJRK-263)

README.md
```diff
@@ -35,7 +35,7 @@ While Infusion allows for just about any possible extension you can imagine and
 | Setting | Description |
 |---------|-------------|
 | `port`  | _(Optional)_ The HTTP port the server will be hosted on. The default value is 8081. If this is changed, it should also be updated in `Dockerfile`, `docker-compose.dev.yml` and `docker-compose.cloud.yml` |
-| `savingEnabled` | Specifies whether saving stories to the database is allowed. If the value is set to `false` and the matching option in `sjrk.storyTelling.baseTheme.page`, then the site is effectively in "read-only" mode and will behave as a collection of stories rather than an authoring tool. For best results, the values in this config file and the `page` grade should match each other. |
+| `authoringEnabled` | Specifies whether creating and saving stories to the database is allowed. If the value is set to `false` and the matching option in `sjrk.storyTelling.base.page`, then the site is effectively in "read-only" mode and will behave as a collection of stories rather than an authoring tool. For best results, the values all `page` grades should be derived from this config file (this is the case in `storyEdit.html`). |
```
cindyli: I think we've decided it's no longer necessary to define the client side `authoringEnabled` flag in `sjrk.storyTelling.base.page`. The client will now fetch this flag from the server.

package.json
```diff
@@ -1,41 +1,43 @@
...
+        "gpii-handlebars": "1.1.0",
+        "handlebars": "4.0.10",
+        "infusion": "3.0.0-dev.20180326T173646Z.8c6a109b1",
+        "kettle": "1.8.0",
+        "sinon": "7.3.2",
+        "uuid": "3.1.0"
+    },
+    "devDependencies": {
+        "chance": "1.0.16",
+        "eslint-config-fluid": "1.1.0",
```
cindyli, line 33: With the use of "gpii-grunt-lint-all", please check if all other dependencies for linting can be removed, such as "eslint-config-fluid", "fluid-grunt-eslint", "grunt-eslint", "grunt-jsonlint". Thanks.

Gruntfile.js
```diff
@@ -15,20 +15,19 @@ module.exports = function (grunt) {
...
         // Project package file destination.
         pkg: grunt.file.readJSON("package.json"),
-        eslint: {
-            all: ["server/src/**/*.js", "ui/src/**/*.js", "tests/**/*.js", "*.js"]
-        },
-        jsonlint: {
-            all: ["package.json", ".eslintrc.json", "server/src/**/*.json", "ui/src/**/*.json", "tests/**/*.json"]
+        lintAll: {
+            sources: {
+                js: ["src/**/*.js", "tests/**/*.js", "*.js"],
```
cindyli: There are also js flies in "themes/" directory. Please add. Thanks.

src/server/requestHandlers.js
```diff
@@ -338,6 +348,31 @@ sjrk.storyTelling.server.deleteSingleFileRecoverable = function (fileToDelete, d
...
+        }
+    }
+});
+
+// Returns a collection of values which are "safe" to share
+// with the client side of the application
+sjrk.storyTelling.server.getClientConfig = function (request, globalConfig, secureConfig) {
+    var clientConfig = {
+        theme: globalConfig.theme || secureConfig.baseThemeName,
+        baseTheme: secureConfig.baseThemeName,
```
cindyli, line 366: As we discussed last Friday, as the client no longer needs to distinguish the base theme from other themes with the change of embedding "base" in component names, the server doesn't need to return `baseTheme: "base"` to the client.

src/server/requestHandlers.js
```diff
@@ -338,6 +348,31 @@ sjrk.storyTelling.server.deleteSingleFileRecoverable = function (fileToDelete, d
...
+        authoringEnabled: globalConfig.authoringEnabled
+    };
+
+    request.events.onSuccess.fire({
+        message: "Client configuration request successfully received",
```
cindyli: Indicating `isError: false` for rejected requests should be enough for the client to distinguish success and rejected response. Returning `clientConfig` object should be good enough: `request.events.onSuccess.fire(clientConfig)`.

src/server/requestHandlers.js
BlueSlug: I kept it in there for the check in `sjrk.storyTelling.loadThemedPage`, since we don't want to load `base.js` or `base.css` (which don't exist) and we need something to compare against in order to make that evaluation.

src/ui/storyTellingServerUI.js
```diff
@@ -0,0 +1,174 @@
+/*
```
cindyli, line 1: It's confusing to have both "server" and "ui" in the file name. Would it be better to rename it to "storyTellingUiHandler" or "storyTellingUiFetcher"?

It also looks redundant to have almost all files in src/ui/ directory starting with "storyTelling" which make file names longer. The project name has been clear about "story-telling".

src/server/requestHandlers.js
cindyli: We should discuss if a header template should be added to handle the inclusion of css and js from theme directories which should also help to solve [SJRK-256 issue](https://issues.fluidproject.org/projects/SJRK/issues/SJRK-256).

tests/server/configs/sjrk.storyTelling.server.testServerWithBaseTheme.json5
```diff
@@ -0,0 +1,19 @@
...
+    "type": "sjrk.storyTelling.server.testServerWithCustomTheme",
+    "options": {
+        "gradeNames": ["sjrk.storyTelling.server"],
+        "components": {
+            "server": {
+                "options": {
+                    "components": {
+                        "baseTheme": {
+                            options: {
+                                root: "./tests/testData/mockBaseTheme"
```
cindyli: Rather than directly overriding `root`, its calculation can be derived by defining configs via globalConfig and secureConfig, which will help to test sjrk.storyTelling.server.getThemePath() function too.

The same comment applies to definitions in tests/server/configs/sjrk.storyTelling.server.testServerWithCustomTheme.json5.

tests/server/testServerWithMissingTheme.js
```diff
@@ -0,0 +1,35 @@
...
+
+/** Configs which can be found to be bad at definition time **/
+
+jqUnit.test("Config with missing theme folder", function () {
+    jqUnit.expectFrameworkDiagnostic("Config with missing theme folder failed as expected", function () {
+        kettle.config.loadConfig({
+            configName: "sjrk.storyTelling.server.testServerWithMissingTheme",
+            configPath: "./tests/server/configs"
+        });
+    }, "The custom theme folder");
```
cindyli: Can you please verify a more meaningful error message? Thanks.

tests/ui/js/storyTellingServerUITests.js
cindyli: Can you track these tests-in-progress in a separate branch so we can merge clean code into the master. Thanks.

cindyli:
Running `grunt lint` reports this warning:

```
Running "wrapped-eslint:md" (wrapped-eslint) task
Could not find any files to validate.
```

README.md
BlueSlug: The value that is ultimately used is the one passed into from the server, but there are still some bits of the templates that are controlled by that flag, as well as at least one IoC reference to it in `storyEdit`, so I've kept it in the `page` grade intentionally. I also intentionally left the reference in the documentation, though it makes sense to remove it from there at this point.

src/server/requestHandlers.js
BlueSlug: That sounds like a good idea, though I feel it might be a bit out of scope for this Jira. I agree that we should talk about it for sure :)

src/ui/storyTellingServerUI.js
BlueSlug: I agree on all counts. The files in the `ui` directory were named roughly according to their grade names in order to help make finding the relevant files easier, though this naming convention is out of keeping with the files in the `server` directory. This is, more than anything else, simply a remnant of the way we kicked off the project with a few days of quick-and-dirty development and a fair amount of premature optimization. I will remove "storyTelling." from all file names and think of a more appropriate name for `storyTellingServerUI.js`.

BlueSlug: In a related issue, I think it would make sense at some point to review all grade names and make sure they still make sense, given that the project has changed so much in the last two years

README.md
cindyli: Why cannot them use the `authoringEnabled` flag passed into from the server as all page components are instantiated with this server flag passed in in their options?

src/server/requestHandlers.js
cindyli: I agree this can be addressed in a separate jira. Can you create one and link with SJRK-256? Thanks.

src/ui/storyTellingServerUI.js
BlueSlug: Re `storyTellingServerUI.js`, I think it might need more than just a renaming. It might make more sense to refactor these functions a bit and put them in other files. `loadBrowse` and `loadStoryFromParameter`, in addition to possibly being renamed themselves, would more logically live in the same file as the `storyBrowse` and `storyView` grades they end up calling. Also my usage of promises and callback parameters simultaneously isn't ideal, so this might be a good opportunity to switch to an all-promises setup.

cindyli: What a coincident that I've been thinking about refactoring of this file too. I actually think all functions in this file can be moved into components and all workflows can be incorporated and controlled by individual components. Use "Browse story" as an example:

1. move the fetch of "/clientConfig" into "sjrk.storyTelling.base.page" `onCreate` event;
2. The retrieval of stories to "/stories" endpoint can be moved to  "sjrk.storyTelling.base.page.storyBrowse" component, in parallel with the fetch of "/clientConfig". It also fires an event "onStoriesReady" when stories are received.
3. The aggregate event for`onClientConfigReady + onStoriesReady` controls other parts such as the load of custom theme files, the instantiation of templateManager and the page rendering.

Considering our timeline, it might be better to do this improvement in a separate JIRA if we decide to do it.

src/server/requestHandlers.js
BlueSlug: [SJRK-264](https://issues.fluidproject.org/browse/SJRK-264) has been filed.

src/ui/storyTellingServerUI.js
BlueSlug: [SJRK-265](https://issues.fluidproject.org/browse/SJRK-265) filed to address this

BlueSlug:
@cindyli ready for review :)

README.md
```diff
@@ -1,140 +1,176 @@
...
+
+#### Provided themes
+
+The "base" files included in this repository, located in the `themes/base` directory, are intended to provide a
+bare-bones implementation of the Storytelling Tool without any project- or organization-specific branding or other
+bells and whistles. The interface is extensible and customizable to enable the creation of new experiences, themes and
+story contexts within the tool.
+The site comes with two custom themes already available: `karisma` and `learningReflections`.
+
+* `learningReflections` Has a few extra pages with content and is styled to illustrate the support of the Storytelling
```
cindyli: Has -> has

src/ui/base-page.js
```diff
@@ -11,19 +11,19 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-    fluid.defaults("sjrk.storyTelling.page", {
+    fluid.defaults("sjrk.storyTelling.base.page", {
         gradeNames: ["fluid.modelComponent"],
         model: {
             uiLanguage: "en" //initial state is English
         },
         pageSetup: {
             resourcePrefix: "",
-            savingEnabled: true // should match the setting in sjrk.storyTelling.server.server.globalConfig
+            authoringEnabled: true // should match the setting in sjrk.storyTelling.server.config.json5
```
cindyli: The current comment is wrong. Please move this line into a commented section and explain it must be provided by integrators and what it is for.

tests/ui/js/storyTellingServerUITests.js
cindyli: bump

BlueSlug:
@cindyli ready for review

tests/ui/js/storyTellingServerUITests.js
cindyli: bump again. Please remove the commented code and track it in your own branch so we can merge the clean code. Thanks.

PR #35, opened 2019-08-05
SJRK-255 - re-review pull to highlight suggestions for refactoring

amb26:


README.md
```diff
@@ -1 +1,207 @@
...
+* Run a CouchDB server on `localhost:5984`
+  * Docker is an easy way to do this, we recommend the `apache/couchdb` image. To use that image via Docker, run this
+    command: `docker run -p 5984:5984 -d apache/couchdb`
+* Run `node .\src\server\db\dbSetup.js` to configure necessary CouchDB databases (this will also ensure your CouchDB
+  instance is set up in [single-node mode](https://docs.couchdb.org/en/stable/setup/single-node.html))
+  * If you have an [admin user](https://docs.couchdb.org/en/stable/intro/security.html) configured on you CouchDB
+  instance, you will need to provide the credentials to `dbSetup.js`. This can be done by setting the `COUCHDB_URL`
+  environment variable. `dbSetup.js` will use the URL specified in `COUCHDB_URL` when connecting to CouchDB. For
+  example: `COUCHDB_URL=http://admin:adminpassword@localhost:5984`
+* Create a `sjrk.storyTelling.server.config.json5` file in the style of `sjrk.storyTelling.server.config.json5.example`
```
amb26, line 37: Be a little more explicit with this instruction and also the one about "secrets" that this can be done by simply copying the existing files - provide the cp command line to eliminate all doubt in the reader's mind!

src/ui/base-page-storyEdit.js
```diff
@@ -0,0 +1,299 @@
...
+                            namespace: "showServerResponse"
+                        }]
+                    },
+                    invokers: {
+                        setShareButtonDisabled: {
+                            this: "{that}.dom.storyShare",
+                            method: "prop",
+                            args: ["disabled", "{arguments}.0"]
+                        },
+                        enableShareButton: {
```
amb26: Eliminate these 6 invokers and lots of the calls to them in favour of mapping this state in model areas - e.g.
progressArea.visible
shareButton.disabled
responseArea.visible
setShareButtonDisabled is an odd man out and should instead be a listener to the change in model state

src/ui/base-page.js
```diff
@@ -0,0 +1,259 @@
...
+                funcName: "sjrk.storyTelling.base.page.renderAllUiTemplates",
+                args: ["{that}"],
+                namespace: "renderAllUiTemplates"
+            },
+            {
+                funcName: "fluid.set",
+                args: ["{uio}", "options.multilingualSettings.locale", "{change}.value"],
+                namespace: "updateUioLanguage"
+            }],
+            "*": {
```
amb26: You probably mean "" here - the "*" syntax means that you want a separate call to set for every individual change to a top-level path - https://docs.fluidproject.org/infusion/development/ChangeApplierAPI.html#wildcards-in-model-path-references

src/ui/base-page.js
```diff
@@ -0,0 +1,259 @@
...
+            },
+            uio: {
+                type: "fluid.uiOptions.prefsEditor.multilingualDemo",
+                container: ".flc-prefsEditor-separatedPanel",
+                options: {
+                    components: {
+                        prefsEditorLoader: {
+                            options: {
+                                components: {
+                                    prefsEditor: {
```
amb26: This is nested inconveniently deep. Instead factor off a prefsEditorInBasePage grade containing these listeners and send the grade down via an options distribution

src/ui/base-page.js
```diff
@@ -0,0 +1,259 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    });
+
+    sjrk.storyTelling.base.page.renderAllUiTemplates = function (component) {
+        fluid.each(component, function (subcomponent) {
```
amb26: This is messy, instead create a renderAllUITemplates event in the component and distribute down an interaction grade to the ui components which listens to is
e.g. 
````
fluid.defaults("sjrk.storyTelling.uiInBasePage", {
    listeners: {
        "{sjrk.storyTelling.base.page}.events.renderAllUiTemplates": "templateManager.events.onResourceLoadRequested.fire"
    }
}
````
and then in the base page
````
distributeOptions: {
    target: "{that sjrk.storyTelling.ui}.options.gradeNames",
    record: "sjrk.storyTelling.uiInBasePage"
    }
````

File: src/ui/storyTellingServerUI.js [hunk 6]
```diff
@@ -0,0 +1,174 @@
...
+        options = options || {};
+        options.distributeOptions = [{
+            "target": "{that storyBrowser}.options.model",
+            "record": browseResponse
+        },{
+            "target": "{that}.options.pageSetup.authoringEnabled",
+            "record": clientConfig.authoringEnabled
+        }];
+
+        var storyBrowseComponent = sjrk.storyTelling[clientConfig.theme].page.storyBrowse(options);
```
amb26, line 100: A big problem with the test set up is that this entire component tree leaks between tests, including the UIO instance nested in it. We need to arrange for teardown of the browse component once the fixtures finish. See below -

tests/ui/js/storyTellingServerUITests.js
```diff
@@ -0,0 +1,213 @@
...
+                },{
+                    // test the CSS/JS injection function directly
+                    funcName: "sjrk.storyTelling.loadCustomThemeFiles",
+                    args: ["sjrk.storyTelling.testUtils.callbackVerificationFunction", {"theme": "learningReflections"}]
+                },{
+                    funcName: "sjrk.storyTelling.storyTellingServerUiTester.assertCustomCssLoaded",
+                    args: ["learningReflections.css", 2]
+                },{
+                    // load a base Browse page and check that the stories are present
+                    task: "sjrk.storyTelling.loadBrowse",
```
amb26: Any time there's a call to loadBrowse we need to arrange for cleanup - I suggest packaging the assertion functions into a reusable utility that as well as making the assertion calls storyBrowse.destroy() once it is done

amb26: A slightly cleaner structure would just have one testEnvironment per use of a "page" and then repackage sjrk.storyTelling.loadThemedPage rather than as a standalone function, a definition for a "createOnEvent" subcomponent of a PageLoader component which fires on the completion of the I/O for /stories (and /clientConfig, with boiling, etc.). Then you could rely on the standard teardown of the testEnvironment to arrange for destruction of the PageLoader once you were finished tinkering with it in the fixtures. This might be enough in itself to stop the race on constructing the 2nd and further UIO components into the same page but I will continue to investigate

src/ui/base-page.js
```diff
@@ -0,0 +1,259 @@
...
+            pageComponent.applier.change("", response);
+            pageComponent.events.onPreferencesLoaded.fire();
+        }, function (error) {
+            pageComponent.events.onPreferenceLoadFailed.fire(error);
+        });
+    };
+
+    sjrk.storyTelling.base.page.reloadUioMessages = function (lang, uioMessageLoaderComponent, uioMessageLoaderLocalePath) {
+        // Set the language in the resource loader
+        fluid.set(uioMessageLoaderComponent, uioMessageLoaderLocalePath, lang);
```
amb26: This function is the source of the race which is causing the tests to fail. Firstly this line updates the component's top-level options which are notionally immutable - then the loadResources line starts a race against the very old-fashioned framework's localisation code and the existing I/O that is already pending at startup. My best feeling at the moment is that it will be less work to update you to my in-progress branch where all of the resource loading code has been rewritten rather than get to the bottom of the inherently race-prone mutation-fest that we have inside old-style fluid.fetchResources. I've got a few fixes to make first before my branch is a drop-in replacement but there should be something testable by tomorrow.

src/ui/UIOMultilingualDemo.js
```diff
@@ -0,0 +1,74 @@
...
+        distributeOptions: {
+            "messageLoaderLocale": {
+                target: "{that messageLoader}.options.model",
+                record: {
+                    resourceLoader: {
+                        locale: "{multilingualDemo}.model.locale"
+                    }
+                }
+            },
+            "relayOnCreateToc": {
```
cindyli, line 39: "onCreateTOC" should be fired when "{messageLoader}.events.onResourcesLoaded" is fired.

But this doesn't solve the problem that the toc header is showing "[Object Object]" when the initial language is Spanish.

src/ui/base-page.js
```diff
@@ -0,0 +1,155 @@
...
+            //authoringEnabled: true
+        },
+        distributeOptions: {
+            "ui.templateManager.authoringEnabled": {
+                source: "{that}.options.pageSetup.authoringEnabled",
+                target: "{that ui templateManager}.options.model.dynamicValues.authoringEnabled"
+            },
+            "ui.templateManager.resourcePrefix": {
+                source: "{that}.options.pageSetup.resourcePrefix",
+                target: "{that ui}.options.components.templateManager.options.templateConfig.resourcePrefix"
```
amb26: This and the next target selector are overlong and fragile - in general it's a mistake to distribute to a tail path that is not attached to the same component. Absorb these into the component selector.

src/ui/UIOMultilingualDemo.js
```diff
@@ -0,0 +1,74 @@
...
+                record: {
+                    "{messageLoader}.events.onResourcesLoaded": [{
+                        func: "{separatedPanel}.events.onCreateSlidingPanelReady",
+                        priority: "before:updateMessageBases",
+                        namespace: "recreateSlidingPanel"
+                    },
+                    {
+                        func: "{prefsEditorLoader}.events.onReady.fire",
+                        priority: "after:recreateSlidingPanel",
+                        namespace: "onslidingPanelReady"
```
amb26: Probably want to normalise case to onSlidingPanelReady

src/ui/UIOMultilingualDemo.js
cindyli: This review comment can be ignored because "onCreateTOCReady", the event that controls the creation of TOC, is [an aggregate event](https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/Enactors.js#L482-L487) including "onMessagesLoaded" so the timing of when "onCreateTOC" is fired becomes less sensitive.

tests/ui/js/storyTellingServerUITests.js
BlueSlug: I have started this work but have found it to be rather complex and outside of the scope of SJRK-255. I started implementing a `pageLoader` grade that used a `fluid.dataSource.URL` to load in the page data, but I ran into a roadblock while attempting to make the `url` option of the DataSource configurable.

This roadblock has been captured in FLUID-6449:
https://issues.fluidproject.org/browse/FLUID-6449

I have also filed a fresh Jira that copies these review comments and the one above them:
https://issues.fluidproject.org/browse/SJRK-320

and my preliminary work on this can be found in this branch:
https://github.com/BlueSlug/sjrk-story-telling/tree/SJRK-320

BlueSlug:
Ready for more review

cindyli:
With the removal of some fields from the document structure for example at [here](https://github.com/fluid-project/sjrk-story-telling/compare/stories-floe-dev...BlueSlug:SJRK-255#diff-d3afd9526f1b047c0728f8dfb5102f63L119-L122), please adjust or add proper comments in tests as these fields are still in use in `testServerWithStorage.js` and `storyFabricator.js`.

BlueSlug:
I've removed the commented-out fields and left a comment detailing them in SJRK-132: https://issues.fluidproject.org/browse/SJRK-132?focusedCommentId=40931

Ready for more review!

PR #37, opened 2019-08-17
SJRK-263: Empty blocks are being saved by storySpeaker

BlueSlug:
This update fixes two bugs:
- the story being saved would save empty blocks
- the storySpeaker (text-to-speech) in the story preview would read out empty blocks

### DIFF
--- a/src/ui/base-page-storyEdit.js
+++ b/src/ui/base-page-storyEdit.js
@@ -78,16 +78,6 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             "audio": ["mediaUrl"],
             "video": ["mediaUrl"]
         },
-        modelRelay: {
-            editorStoryToPreviewer: {
-                target: "{storyPreviewer}.story.model",
-                singleTransform: {
-                    type: "fluid.transforms.free",
-                    func: "sjrk.storyTelling.base.page.storyEdit.removeEmptyBlocks",
-                    args: ["{storyEditor}.story.model", "{that}.options.blockContentValues"]
-                }
-            }
-        },
         components: {
             storySpeaker: {
                 options: {
@@ -215,6 +205,21 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
                                     }
                                 }
                             }
+                        },
+                        story: {
+                            options: {
+                                model: "{storyEditor}.story.model",
+                                modelRelay: {
+                                    contentEmptyBlockFilter: {
+                                        target: "content",
+                                        singleTransform: {
+                                            type: "fluid.transforms.free",
+                                            func: "sjrk.storyTelling.base.page.storyEdit.removeEmptyBlocks",
+                                            args: ["{storyPreviewer}.story.model.content", "{storyEdit}.options.blockContentValues"]
+                                        }
+                                    }
+                                }
+                            }
                         }
                     }
                 }
@@ -222,24 +227,25 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         }
     });
 
-    /* Removes all empty blocks from a given array of story blocks
-     * - "blocks": an array of story blocks
-     * - "blockContentValues": a collection of arrays which outline the values
+    /* Removes all empty blocks from a given collection of story blocks
+     * - "blocks": a collection of story blocks (sjrk.storyTelling.block)
+     * - "blockContentValues": a hash map of block types which outlines the values
      *      that, if at least one is truthy, means a particular block is not empty
      */
-    sjrk.storyTelling.base.page.storyEdit.removeEmptyBlocks = function (storyModel, blockContentValues) {
-        storyModel.content = fluid.remove_if(storyModel.content, function (block) {
+    sjrk.storyTelling.base.page.storyEdit.removeEmptyBlocks = function (blocks, blockContentValues) {
+        var filteredBlocks = fluid.remove_if(blocks, function (block) {
             return sjrk.storyTelling.base.page.storyEdit.isEmptyBlock(block, blockContentValues[block.blockType]);
         });
 
-        return storyModel;
+        return filteredBlocks;
     };
 
     /* Returns true if a block is determined to be empty, based on the values
      * listed in blockContentValues. If at least one of those values is truthy,
      * the block is not empty.
-     * - "block": a story block
-     * - "blockContentValues": an array of the values as described above
+     * - "block": a single story block (sjrk.storyTelling.block)
+     * - "blockContentValues": a hash map of block types which outlines the values
+     *      that, if at least one is truthy, means a particular block is not empty
      */
     sjrk.storyTelling.base.page.storyEdit.isEmptyBlock = function (block, blockContentValues) {
         return !fluid.find_if(blockContentValues, function (blockContentValues) {
--- a/src/ui/base-page-storyEdit.js
+++ b/src/ui/base-page-storyEdit.js
@@ -56,7 +56,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             },
             "onStoryShareRequested.submitStory": {
                 funcName: "sjrk.storyTelling.base.page.storyEdit.submitStory",
-                args: ["{storyEditor}", "{that}.events.onStoryShareComplete"]
+                args: ["{storyPreviewer}", "{that}.events.onStoryShareComplete"]
             },
             "onCreate.setEditorDisplay": {
                 func: "{that}.setEditorDisplay"
--- a/src/ui/base-page-storyEdit.js
+++ b/src/ui/base-page-storyEdit.js
@@ -245,15 +245,16 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     };
 
     /* Returns true if a block is determined to be empty, based on the values
-     * listed in blockContentValues. If at least one of those values is truthy,
-     * the block is not empty.
+     * listed in blockContentValuesForType. If at least one of those values is
+     * truthy, the block is not empty. If the values are empty or otherwise can't
+     * be iterated over, then the block is also empty regardless of its contents.
      * - "block": a single story block (sjrk.storyTelling.block)
-     * - "blockContentValues": a hash map of block types which outlines the values
-     *      that, if at least one is truthy, means a particular block is not empty
+     * - "blockContentValuesForType": an array of values for the block's type
+     *      that, if at least one is truthy, mean the block is not empty
      */
-    sjrk.storyTelling.base.page.storyEdit.isEmptyBlock = function (block, blockContentValues) {
-        return !fluid.find_if(blockContentValues, function (blockContentValues) {
-            return !!block[blockContentValues];
+    sjrk.storyTelling.base.page.storyEdit.isEmptyBlock = function (block, blockContentValuesForType) {
+        return !fluid.find_if(blockContentValuesForType, function (blockContentValue) {
+            return !!block[blockContentValue];
         });
     };
 
--- a/src/ui/base-page-storyEdit.js
+++ b/src/ui/base-page-storyEdit.js
@@ -233,8 +233,12 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
      *      that, if at least one is truthy, means a particular block is not empty
      */
     sjrk.storyTelling.base.page.storyEdit.removeEmptyBlocks = function (blocks, blockContentValues) {
-        var filteredBlocks = fluid.remove_if(blocks, function (block) {
-            return sjrk.storyTelling.base.page.storyEdit.isEmptyBlock(block, blockContentValues[block.blockType]);
+        var filteredBlocks = [];
+
+        fluid.each(blocks, function (block) {
+            if (!sjrk.storyTelling.base.page.storyEdit.isEmptyBlock(block, blockContentValues[block.blockType])) {
+              filteredBlocks.push(block);
+            }
         });
 
         return filteredBlocks;
--- a/src/ui/base-page-storyEdit.js
+++ b/src/ui/base-page-storyEdit.js
@@ -56,7 +56,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             },
             "onStoryShareRequested.submitStory": {
                 funcName: "sjrk.storyTelling.base.page.storyEdit.submitStory",
-                args: ["{storyEditor}", "{storyPreviewer}.story.model", "{that}.events.onStoryShareComplete"]
+                args: ["{storyEditor}.dom.storyEditorForm", "{storyPreviewer}.story.model", "{that}.events.onStoryShareComplete"]
             },
             "onCreate.setEditorDisplay": {
                 func: "{that}.setEditorDisplay"
@@ -237,7 +237,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
 
         fluid.each(blocks, function (block) {
             if (!sjrk.storyTelling.base.page.storyEdit.isEmptyBlock(block, blockContentValues[block.blockType])) {
-              filteredBlocks.push(block);
+                filteredBlocks.push(block);
             }
         });
 
@@ -262,16 +262,14 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         $(pageContainer).toggleClass(hiddenEditorClass, !authoringEnabled);
     };
 
-    sjrk.storyTelling.base.page.storyEdit.submitStory = function (storyEditor, storyModel, errorEvent) {
-        var form = storyEditor.container.find("form");
-
-        form.attr("action", "/stories/");
-        form.attr("method", "post");
-        form.attr("enctype", "multipart/form-data");
+    sjrk.storyTelling.base.page.storyEdit.submitStory = function (storyEditorForm, storyModel, errorEvent) {
+        storyEditorForm.attr("action", "/stories/");
+        storyEditorForm.attr("method", "post");
+        storyEditorForm.attr("enctype", "multipart/form-data");
 
         // This is the easiest way to be able to submit form
         // content in the background via ajax
-        var formData = new FormData(form[0]);
+        var formData = new FormData(storyEditorForm[0]);
 
         // Stores the entire model as a JSON string in one
         // field of the multipart form
@@ -282,8 +280,8 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         // proper handling of feedback on success / failure,
         // but currently it just logs to console
         $.ajax({
-            url         : form.attr("action"),
-            data        : formData ? formData : form.serialize(),
+            url         : storyEditorForm.attr("action"),
+            data        : formData || storyEditorForm.serialize(),
             cache       : false,
             contentType : false,
             processData : false,
@@ -297,11 +295,8 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             error       : function (jqXHR, textStatus, errorThrown) {
                 fluid.log("Something went wrong");
                 fluid.log(jqXHR, textStatus, errorThrown);
-                var errorMessage = "Internal server error";
-                if (jqXHR && jqXHR.responseJSON && jqXHR.responseJSON.message) {
-                    errorMessage = jqXHR.responseJSON.message;
-                }
-                errorEvent.fire(errorMessage);
+
+                errorEvent.fire(fluid.get(jqXHR, ["responseJSON", "message"]) || "Internal server error");
             }
         });
     };
--- a/src/ui/base-page-storyEdit.js
+++ b/src/ui/base-page-storyEdit.js
@@ -56,7 +56,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             },
             "onStoryShareRequested.submitStory": {
                 funcName: "sjrk.storyTelling.base.page.storyEdit.submitStory",
-                args: ["{storyPreviewer}", "{that}.events.onStoryShareComplete"]
+                args: ["{storyEditor}", "{storyPreviewer}.story.model", "{that}.events.onStoryShareComplete"]
             },
             "onCreate.setEditorDisplay": {
                 func: "{that}.setEditorDisplay"
@@ -258,8 +258,8 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         $(pageContainer).toggleClass(hiddenEditorClass, !authoringEnabled);
     };
 
-    sjrk.storyTelling.base.page.storyEdit.submitStory = function (that, errorEvent) {
-        var form = that.container.find("form");
+    sjrk.storyTelling.base.page.storyEdit.submitStory = function (storyEditor, storyModel, errorEvent) {
+        var form = storyEditor.container.find("form");
 
         form.attr("action", "/stories/");
         form.attr("method", "post");
@@ -271,7 +271,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
 
         // Stores the entire model as a JSON string in one
         // field of the multipart form
-        var modelAsJSON = JSON.stringify(that.story.model);
+        var modelAsJSON = JSON.stringify(storyModel);
         formData.append("model", modelAsJSON);
 
         // In the real implementation, this should have
--- a/src/ui/base-page-storyEdit.js
+++ b/src/ui/base-page-storyEdit.js
@@ -264,9 +264,11 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     };
 
     sjrk.storyTelling.base.page.storyEdit.submitStory = function (storyEditorForm, storyModel, errorEvent) {
-        storyEditorForm.attr("action", "/stories/");
-        storyEditorForm.attr("method", "post");
-        storyEditorForm.attr("enctype", "multipart/form-data");
+        storyEditorForm.attr({
+            action: "/stories/",
+            method: "post",
+            enctype: "multipart/form-data"
+        });
 
         // This is the easiest way to be able to submit form
         // content in the background via ajax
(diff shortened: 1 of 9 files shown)


jobara:
@BlueSlug I'm having trouble actually running this locally. The site loads, but the tool is missing from the storyEdit.html page and the storiesBrowse.html page throws an error related to the options distributions. 

> "Assertion failure - check console for more details: Error in options distribution record {"target":"{that}.options.pageSetup.authoringEnabled","namespace":"no-namespace-distribution-jtkpu25-155"}: must supply either a member "source" holding an IoC reference or a member "record" holding a literal record"

src/ui/base-page-storyEdit.js
```diff
@@ -215,31 +205,47 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-     * - "blockContentValues": a collection of arrays which outline the values
+    /* Removes all empty blocks from a given collection of story blocks
+     * - "blocks": a collection of story blocks (sjrk.storyTelling.block)
+     * - "blockContentValues": a hash map of block types which outlines the values
      *      that, if at least one is truthy, means a particular block is not empty
      */
-    sjrk.storyTelling.base.page.storyEdit.removeEmptyBlocks = function (storyModel, blockContentValues) {
-        storyModel.content = fluid.remove_if(storyModel.content, function (block) {
+    sjrk.storyTelling.base.page.storyEdit.removeEmptyBlocks = function (blocks, blockContentValues) {
+        var filteredBlocks = fluid.remove_if(blocks, function (block) {
```
jobara: `[fluid.remove_if](https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/Fluid.js#L622-L656)` modifies the source. In this case the code will modify the `blocks` in place, as well as returning the `filteredBlocks` which will have the same value. I think you want to do a `fluid.copy` on blocks and pass that in to the `fluid.remove_if` function. Although it might be better just to use `fluid.each`, `fluid.accumulate`, or something like that just to iterate over the source and generate a filtered source with just the values wanted.

src/ui/story.js
```diff
@@ -47,24 +47,26 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-                input: "{that}.model.content",
-                separator: ". ",
-                stringOnly: true,
-                path: "contentString"
+        modelRelay: {
+            contentToContentString: {
+                target: "contentString",
+                singleTransform: {
+                    type: "sjrk.storyTelling.transforms.arrayToString",
+                    input: "{that}.model.content",
```
jobara: Instead of specifying `input`, is it possible to just add this as the modelRelay `source`?

src/ui/transforms.js
```diff
@@ -58,22 +58,22 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-        var path = extraInputs.path();
+        var combinedString = "";
+        var separator = extraInputs.separator(),
+            stringOnly = extraInputs.stringOnly(),
+            path = extraInputs.path();
 
         fluid.each(input, function (term) {
             if (path) {
                 term = fluid.get(term, path) || term;
             }
```
jobara: It looks like `path` defaults to "", can you write this whole block as `term = fluid.get(term, path)`? Also it looks like `stringOnly` defaults to `""` but should probably default to `false`.

src/ui/transforms.js
```diff
@@ -58,22 +58,22 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
 
             if (!stringOnly || (term && typeof term === "string")) {
```
jobara: Do you need the guard on `term`? I think `typeof(term)` would return `"undefined"` if it didn't exist. Or, are you trying to prevent adding empty strings?

jobara:
@BlueSlug I'm also having trouble running the tests. I've tried both through my own web server and the SJRK web server. I wonder if I missed a step getting things setup to run?

BlueSlug:
> @BlueSlug I'm having trouble actually running this locally. The site loads, but the tool is missing from the storyEdit.html page and the storiesBrowse.html page throws an error related to the options distributions.
> 
> > "Assertion failure - check console for more details: Error in options distribution record {"target":"{that}.options.pageSetup.authoringEnabled","namespace":"no-namespace-distribution-jtkpu25-155"}: must supply either a member "source" holding an IoC reference or a member "record" holding a literal record"

as we discussed on IRC, it seems your configuration file was out of date and missing the `authoringEnabled` key. It still had the old `savingEnabled` name, so the options distribution didn't have a value to pass along.

src/ui/transforms.js
BlueSlug: Correct, I added that to keep empty strings from being added to the resulting string.

BlueSlug:
@jobara I made a couple of the changes you recommended, except for the guard on `term` (`transforms.js:69`). I also noticed that the tests for transforms.js were woefully out of date, so I updated them as well.

Ready for more review :)

src/ui/base-page-storyEdit.js
```diff
@@ -215,31 +205,51 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
     };
 
     /* Returns true if a block is determined to be empty, based on the values
      * listed in blockContentValues. If at least one of those values is truthy,
      * the block is not empty.
-     * - "block": a story block
-     * - "blockContentValues": an array of the values as described above
+     * - "block": a single story block (sjrk.storyTelling.block)
+     * - "blockContentValues": a hash map of block types which outlines the values
+     *      that, if at least one is truthy, means a particular block is not empty
```
jobara: I'm debating in my head if this is the right check or not. It seems that blocks would only ever hold strings. That is text or urls. Is that correct? Should other values be stripped. That is, would you accept a Number, Array, Object, etc. Also, it would accept a string only made of whitespace, or invalid URLs. In terms of users interacting through the UI, I think you'd only really have to worry about strings only made of whitespace, but it's probably still worth investigating how you handle the other conditions as well.

src/ui/base-page-storyEdit.js
```diff
@@ -252,8 +262,8 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         $(pageContainer).toggleClass(hiddenEditorClass, !authoringEnabled);
     };
 
-    sjrk.storyTelling.base.page.storyEdit.submitStory = function (that, errorEvent) {
-        var form = that.container.find("form");
+    sjrk.storyTelling.base.page.storyEdit.submitStory = function (storyEditor, storyModel, errorEvent) {
+        var form = storyEditor.container.find("form");
```
jobara: Can you use the `storyEditorForm` selector instead of `find`? Also I couldn't leave a comment there, but here are some suggestions for the ajax call. 

1. for the data property you should be able to use `formData || form.serialize()`
2. the error message could be assigned like `var errorMessage = fluid.get(jqXHR, ["responseJSON", "message"]) || "Internal server error";` You may also want to put these messages in the `strings` block of the component.
3. Could you use [DataSource](https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/DataSource.js) for this?

src/ui/transforms.js
```diff
@@ -40,40 +40,42 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
     fluid.defaults("sjrk.storyTelling.transforms.arrayToString", {
         "gradeNames": [ "fluid.standardTransformFunction", "fluid.multiInputTransformFunction" ],
         "inputVariables": {
             "separator": ", ",
-            "stringOnly": "",
+            "trailingSeparator": false,
+            "stringOnly": true,
             "path": ""
         }
     });
```
jobara, line 59: You might as well add an inverse method as well. And can you file a JIRA about contributing these up to Infusion. A contributed version may need to be generalized a bit though.

src/ui/transforms.js
```diff
@@ -40,40 +40,42 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
             if (!stringOnly || (term && typeof term === "string")) {
-                contentString += term + (separator || "");
+                combinedString += term + (separator || "");
             }
         });
 
-        return contentString.substring(0, contentString.lastIndexOf(separator));
+        if (!trailingSeparator) {
+            combinedString = combinedString.substring(0, combinedString.lastIndexOf(separator));
+        }
```
jobara: It might be easier to just push acceptable strings into an array and calling the [join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) method.

tests/ui/js/transformsTests.js
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 OCAD University
+Copyright 2017-2019 OCAD University
```
jobara, line 2: Is 2017 the correct start date? You may also want to transition this repo, in a future PR,  to the centralized Copyright like we now have in Infusion.

tests/ui/js/transformsTests.js
```diff
@@ -37,24 +37,102 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            rules: {
+                path: "notARealPath"
+            },
+            expectedResult: ""
+        },
+        {
+            sourceArray: ["Shyguy", "Rootbeer", "", false, {}, []],
+            rules: {},
+            expectedResult: "Shyguy, Rootbeer"
+        }
```
jobara, line 251: should probably test if `sourceArray` is empty, an object ( from your implementation it looks like it would actually work), a string, undefined, and etc.

src/ui/base-page-storyEdit.js
BlueSlug: That's a good question. In terms of what is currently accepted, it is the case that it's only going to be strings or URL's, though it's possible that other values could be added for different block types in the future. We had discussed a "drawing" type block at some point, so I can see that the implementation of that would potentially have more complex data stored at a particular point on its model, in which case limiting the values to strings may not go so well.

I agree that checking for all-whitespace strings is a good idea. That might be something left to field validation (currently nonexistent) rather than the filter. What do you think?

src/ui/transforms.js
BlueSlug: Can you elaborate on what you mean by an inverse method? Would the transformation above it (stringToArray) fit the bill, or do you mean something within arrayToString that indicates what its inverse is?

tests/ui/js/transformsTests.js
BlueSlug: 2017 is indeed the correct start date. At one point in 2018, I updated all the copyright dates to 2018 rather than indicate a range, so I've been fixing these here and there as I see them.

src/ui/transforms.js
BlueSlug: This is a good suggestion and I'll explore this a little more. The function was previously literally just a call to .join, but it has since become more complex.

src/ui/transforms.js
jobara: Yes, that function would probably work. You'd need to setup the `invertConfiguration` option in the transformation grades.

src/ui/base-page-storyEdit.js
jobara: It's good to note that at the moment an empty array or object would also pass. 

Regarding validation and filtering, in an ideal case you'd have both. This way the user knows right away that something is invalid, and we also prevent programmatic manipulation of values.

src/ui/transforms.js
```diff
@@ -40,40 +40,42 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-     * - stringOnly (optional): flag to exclude all falsy or non-string values
-     * - path (optional): an EL path on each item in the terms collection
+     * Values can also be accessed via a specific object path relative to each term.
+     * - "separator" (optional): the delimiter to be inserted between each term
+     * - "trailingSeparator" (optional): flag to include the separator at the end
+     * - "stringOnly" (optional): flag to allow only non-empty strings
+     * - "path" (optional): an EL path on each item in the terms collection
      */
     fluid.defaults("sjrk.storyTelling.transforms.arrayToString", {
         "gradeNames": [ "fluid.standardTransformFunction", "fluid.multiInputTransformFunction" ],
```
amb26: This isn't really a multiInputTransformFunction since we don't expect to resolve all of these values from model material dynamically. This can just be a standardTransformFunction with a bunch of options in the transformSpec at top level.

src/ui/transforms.js
amb26: This check is a bit vague and given it is prepared for terms to be nonstrings, it should also be prepared to deal with oddities such as finding the number 0. I suggest checking against checking against the empty string explicitly.

BlueSlug: @amb26 do you mean something like `if (!stringOnly || (typeof term === "string" && term !== ""))`?

By the current logic, if it's the number 0 not wrapped in a string, then it will be excluded if `stringsOnly` is true, and if that option is false then `0` will be allowed. I suppose I could add some checks for falsy or for truthy but meaningless values like:
- the empty array
- the empty object
- `"0"` (the string)
- `"false"` (the string)

I feel that at this point, my goal is to put only the things that convey information into a human-readable string, i.e. something that looks a bit like natural language without all the accoutrement of curly braces and double-quotes. That said, if I intend to merge this into Infusion some day, these restrictions may be too much and we might want to keep it broad and allow falsy or "meaningless" values to be stringified as-is.

I'm also unsure about how to handle complex objects, given `JSON.stringify` will include all of the object structure and layout. This would make full inversion more easily possible, but readability worse.

tests/ui/js/transformsTests.js
BlueSlug: [SJRK-267](https://issues.fluidproject.org/browse/SJRK-267) has been filed to move to a centralized copyright

BlueSlug:
Ready for review. Tests for inverted transforms to come.

src/ui/base-page-storyEdit.js
```diff
@@ -252,28 +263,26 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-    sjrk.storyTelling.base.page.storyEdit.submitStory = function (that, errorEvent) {
-        var form = that.container.find("form");
-
-        form.attr("action", "/stories/");
-        form.attr("method", "post");
-        form.attr("enctype", "multipart/form-data");
+    sjrk.storyTelling.base.page.storyEdit.submitStory = function (storyEditorForm, storyModel, errorEvent) {
+        storyEditorForm.attr("action", "/stories/");
+        storyEditorForm.attr("method", "post");
+        storyEditorForm.attr("enctype", "multipart/form-data");
```
jobara: You can do this in a single `attr` call by using an object as the argument. e.g. 

```json
{
    "action": "/stories/",
    "method": "post",
    "enctype": "multipart/form-data"
}
```
see: https://api.jquery.com/attr/#attr-attributes

src/ui/transforms.js
```diff
@@ -39,41 +43,46 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-        var contentString = "";
-        var separator = extraInputs.separator();
-        var stringOnly = extraInputs.stringOnly();
-        var path = extraInputs.path();
+    sjrk.storyTelling.transforms.arrayToString = function (input, transformSpec) {
+        var delimiter = transformSpec.delimiter || ", ",
+            stringOnly = transformSpec.stringOnly,
+            path = transformSpec.path || "";
+
+        stringOnly = stringOnly === undefined ? true : stringOnly;
```
jobara: If you want to also convert `null` to `true`, you may want to use `fluid.isValue` as the check. Similar for the `trim` option in sjrk.storyTelling.transforms.stringToArray.

src/ui/transforms.js
```diff
@@ -39,41 +43,46 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
 
-            if (!stringOnly || (term && typeof term === "string")) {
-                contentString += term + (separator || "");
+            if (!stringOnly || (term !== "" && typeof term === "string")) {
+                terms.push(term);
             }
         });
 
-        return contentString.substring(0, contentString.lastIndexOf(separator));
+        return terms.join(delimiter || "");
```
jobara: Can `delimiter` ever be falsy? On line 64 you have a falsy check to default to ", "

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -863,6 +879,303 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         return fluid.get(component, "story.model.content.0." + fieldName);
     };
 
+    var isEmptyBlockTestCases = [
```
jobara: Should be attached to the public namespace e.g. `sjrk.storyTelling.base.page.storyEditTester.isEmptyBlockTestCases`.

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -863,6 +879,303 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                heading: "",
+                alternativeText: "",
+                description: "",
+                transcript: "",
+                mediaUrl: "Not really a URL"
+            },                  blockContentValues: "default",  expectedEmpty: false
+        }
+    ];
+
+    sjrk.storyTelling.base.page.storyEditTester.verifyIsEmptyBlock = function (blockContentValues) {
```
jobara: Maybe rename the arg to defaultBlockContentValues or something like that. Actually I wonder if the default should be part of the testCases structure. Something like:

```json
{
    "defaultBlockContentValues": [],
    "testCases": []
}
```

As it stands now, the test cases are assuming a specific default value but these are actually passed in to the tester function when it is called.

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -863,6 +879,303 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+        fluid.each(isEmptyBlockTestCases, function (testCase, index) {
+            var blockContentValuesToUse = testCase.blockContentValues === "default" ? blockContentValues : testCase.blockContentValues;
```
jobara: Instead of having the `"default"` flag, perhaps you can just test if the a value is provided and if not, use the default. Unless you are trying to explicitly test when `blockConentValues` is `undefined` or `null`, but I didn't notice any test cases like that.

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -863,6 +879,303 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+
+            var actuallyEmpty = sjrk.storyTelling.base.page.storyEdit.isEmptyBlock(testCase.block, blockContentValuesToUse[testCase.block.blockType]);
+            jqUnit.assertEquals("Block emptiness state for test case " + index + " is as expected", testCase.expectedEmpty, actuallyEmpty);
```
jobara, line 1176: May want to consider naming the test cases, or providing some more detail in the output string. If one were to break, it would be hard to locate it because of the number test cases provided.

jobara:
I'm seeing a test failure for the Story Page -Tests (base-page-Tests.html). Test page grade: Test events and timing, test #4. 

```
4. Uncaught TypeError: Cannot read property 'replace' of undefined
Source: | http://localhost:8081/node_modules/infusion/dist/infusion-all.js:33502
```

jobara:
In testing out the tool on the site, I added an image block, but didn't upload an Image. I did add a description, which wasn't read out or saved, but also skipped over adding and Title, Author or keywords. I was able to save this to the database. However, it seems that it is essentially empty with none of the model fields having anything more than `""` or `[]` as values. Should we be able to publish this type of story?

tests/ui/js/base-page-storyEditTests.js
BlueSlug: I think putting the default values in the test case in the `blockContentValues` makes the most sense, but I thought passing it in and using `"default"` in the test cases was a better option since we avoid duplicating the structure 30-odd times.

tests/ui/js/base-page-storyEditTests.js
BlueSlug: This would be a good solution to the issue above (whether to provide default values in each test case).

tests/ui/js/base-page-storyEditTests.js
BlueSlug: I agree, it would be easy to lose count while navigating the collection. I had made an attempt at this, and then I found that trying to name 51 test cases meaningfully was difficult. I'll give it another shot :)

BlueSlug:
> I'm seeing a test failure for the Story Page -Tests (base-page-Tests.html). Test page grade: Test events and timing, test #4.
> 
> ```
> 4. Uncaught TypeError: Cannot read property 'replace' of undefined
> Source: | http://localhost:8081/node_modules/infusion/dist/infusion-all.js:33502
> ```

This is symptomatic of the issue described in [SJRK-255](https://issues.fluidproject.org/browse/SJRK-255), and should go away once that is solved. These tests should pass if you run them again, perhaps just that file on its own. The same will occur with `base-page-storyEdit-Tests.html`.

BlueSlug:
> In testing out the tool on the site, I added an image block, but didn't upload an Image. I did add a description, which wasn't read out or saved, but also skipped over adding and Title, Author or keywords. I was able to save this to the database. However, it seems that it is essentially empty with none of the model fields having anything more than `""` or `[]` as values. Should we be able to publish this type of story?

The eventual intent is to disallow the saving of an empty story entirely, but for the time being the behaviour you've described is what we expect.

The Jira that covers disabling the saving of empty stories is [SJRK-196](https://issues.fluidproject.org/browse/SJRK-196)

BlueSlug:
Ready for review

tests/ui/js/base-page-storyEditTests.js
jobara: In the example I gave above the default values is adjacent to all of the test cases, so you can have it grouped together, but only defined once.

src/ui/story.js
```diff
@@ -47,24 +47,28 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                source: "{that}.model.content",
+                target: "contentString",
+                singleTransform: {
+                    type: "sjrk.storyTelling.transforms.arrayToString",
+                    delimiter: ". ",
+                    path: "contentString"
+                },
+                backward: {
+                    excludeSource: "*"
+                }
```
jobara: you can also use `backward: "never"`

src/ui/transforms.js
```diff
@@ -13,24 +13,25 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-        var sourceString = input,
-            delimiter = extraInputs.delimiter(),
-            trim = extraInputs.trim();
+    sjrk.storyTelling.transforms.stringToArray = function (input, transformSpec) {
+        if (!input || typeof input !== "string") {
+            return [];
+        }
+
+        return fluid.transform(input.split(transformSpec.delimiter || ","), function (tag) {
+            var trim = fluid.isValue(transformSpec.trim) ? transformSpec.trim : true;
```
jobara: we can probably improve performance here a bit. First the fluid.transform function is another of those iterator type functions. So we are checking trim for each item in the array. Additionally we don't even need to run the transform if trim is falsy.

jobara:
This is out of scope for this PR, but to make sure I don't forget, have you considered using [JSDoc](https://jsdoc.app) syntax for the code comments?

src/ui/story.js
BlueSlug: Good to know! That isn't in the Infusion docs so I wasn't aware of this option :)

BlueSlug:
> This is out of scope for this PR, but to make sure I don't forget, have you considered using [JSDoc](https://jsdoc.app) syntax for the code comments?

I've seen it in Infusion here and there, and the thought had crossed my mind! I was already trying to add comments in a vaguely unified style for each function in the project, but I'd never made any concerted effort to cover the entire codebase, and it was just a style that I'd made up rather than any formal syntax to be used with an API generator :)

I've filed [SJRK-269](https://issues.fluidproject.org/browse/SJRK-269) to address this.

Also ready for review

src/ui/story.js
```diff
@@ -47,24 +47,26 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-                separator: ". ",
-                stringOnly: true,
-                path: "contentString"
+        modelRelay: {
+            contentToContentString: {
+                source: "{that}.model.content",
+                target: "contentString",
+                singleTransform: {
+                    type: "sjrk.storyTelling.transforms.arrayToString",
+                    delimiter: ". ",
```
jobara, line 56: This is causing an interesting "bug". 

1. Add a text block with text like "Some Text!"
2. Add another text block with text like "more text"
3. The output is "Some Text!. more text"; Notice the double punctuation "!.".

I imagine that what happens with that will depend on the synthesizer used. Some might ignore it, others might read it as "dot". If you end up using the contentString for output the text visually, it will look strange and be incorrect. Also what about other languages that have different punctuation? It may be better not to try to add punctuation and just separate based on " ". What do you think?

BlueSlug: It certainly isn't an ideal situation one way or the other. If I introduce some special case coding to check for grammatical incompatibility with the delimiter, then we're making this transform function less general and therefore less likely to be able to be merged into Infusion in the future. If we don't, then as you've described we risk tripping up certain screen readers or other AT's.

If we set the delimiter to `" "` then we won't get the pause between blocks, and the synthesizer may continue reading the next one as if it's all one sentence:
1. Add a text block with text like "Some text"
2. Add another text block with text like "more text"
3. the output is "Some text more text", which will be read without pauses and could cause confusion

I'm not sure what the best approach is, here. Perhaps we can file a new Jira to capture this issue and have a discussion there, and scope it to make the necessary changes to detect and compensate for punctuation at the end of a term that would be incompatible with the delimiter.

jobara: Sure, file a new JIRA. I'm starting to wonder more if you should be storing the contentString in the model at all. Perhaps even just using the Orator component to read through the content in the story instead. Not sure if this would solve the issue of pausing between "unfinished" blocks, but it might help with other things like the empty "by" and "keywords" sections. Just throwing some ideas out there for the new JIRA.

jobara:
merged at 8f7600cb049323ce088745f42f1733ad73d774fe

src/ui/story.js
BlueSlug: [SJRK-271](https://issues.fluidproject.org/browse/SJRK-271) has been filed to capture this

PR #40, opened 2019-09-09
SJRK-274: Using Orator component from Infusion

jobara:
TTS switched to use the Orator component from Infusion

https://issues.fluidproject.org/browse/SJRK-274


jobara:

themes/base/templates/storyViewer.handlebars
```diff
@@ -6,37 +6,35 @@
...
+            <p class="sjrkc-st-story-author sjrk-st-story-author">{{localizedMessages.message_authorText}}</p>
+        {{/if}}
 
-    <div hidden class="sjrk-st-language-container">
-        {{localizedMessages.message_languageLabelText}}: <span class="sjrkc-st-story-language sjrk-st-story-language">
-            {{lookup ../localizedMessages.availableLanguages dynamicValues.story.language}}
-        </span>
-    </div>
+        <div hidden class="sjrk-st-link-container">
+            <a class="sjrkc-st-story-read-more sjrk-st-story-read-more" href="#">{{localizedMessages.message_readMoreText}}</a>
```
cindyli: What is this `read-more` container for? It doesn't seem being used anywhere.

cindyli:
It would be nice to add some tests for the switch to the orator component.

themes/base/templates/storyViewer.handlebars
jobara: @cindyli I'm not sure. This was already present in the file. Perhaps @BlueSlug would know.

jobara:
@cindyli I've added some tests for the Orator. Ready for more review.

themes/base/templates/storyViewer.handlebars
BlueSlug: @cindyli @jobara it's one of many bits of markup leftover from one of the original designs for the tool. The intention was to have a link users could click to see other stories similar to the one they are currently viewing. If/when we implement tagging, this would take advantage of that.

I can file a Jira to remove all such pieces, and in doing so preserve them

cindyli:
On the preview page, when the "play" button shows up for selected text, the button position is not correct. This doesn't happen on the "view story" page.

Screen Shot 2019-09-11 at 12 03 58 PM

jobara:
@cindyli that's an interesting find. I'm not really sure what's going on with this specific page. I wonder if we should just disable the "selected text" TTS for now. It wasn't in the original version anyways. Thoughts? @BlueSlug do you have any thoughts too?

cindyli:
@jobara I'm fine with disabling the "selected text" TTS if it's not in the original design.

jobara:
I think the issue is related to some of the relative positioning of elements above or around the orator play button. Unfortunately I don't think we'll have time to sort through all of that at the moment, so I'll just disable the selected text feature for the time being.

jobara:
I've filed a new JIRA for addressing the positioning in the future. https://issues.fluidproject.org/browse/SJRK-283

cindyli:
Merged at 25367c2b0b56261763e854b698ed028e27a16e62

PR #41, opened 2019-09-11
SJRK-273: Resolve identified a11y issues for Cities Co-design session

BlueSlug:
This pull request solves a number of issues as identified in the [SJRK-273 Epic Jira](https://issues.fluidproject.org/browse/SJRK-273)

@cindyli to review :)

Once this has been merged in, it should then be merged into the `stories-cities-production` branch as well


src/ui/block-timeBased.js
```diff
@@ -17,15 +17,13 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         model: {
             mediaUrl: null,
             alternativeText: null,
-            description: null,
-            transcript: null,
-            hasTranscript: true
```
cindyli, line 22: The value of `hasTranscript` is still expected in `themes/base/templates/storyBlockMediaEdit.handlebars` file.

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -881,44 +838,24 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
         },
         "test_30": { expectedEmpty: false,
             block: {
                 blockType: "text",
                 heading: "An actual heading",
-                text: "",
-                simplifiedText: ""
-            }
-        },
-        "test_31": { expectedEmpty: false,
```
cindyli, line 904: It might be useful to rename sequential numbers in keys with short meaningful test case purpose to: 1. provide a little explanation about the test case; 2. reduce the tediousness of adjusting the numbers when one or more test cases are removed.

tests/ui/js/block-timeBasedTests.js
```diff
@@ -87,70 +86,6 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-                {
-                    func: "{timeBased}.applier.change",
-                    args: ["transcript", "Rootbeer: krrraow"]
-                },
-                {
-                    funcName: "jqUnit.assertEquals", // 0001
-                    args: ["Combined text is as expected", "Rootbeer: krrraow", "{timeBased}.model.contentString"]
-                },
-                {
-                    func: "{timeBased}.applier.change",
```
cindyli, line 100: What gets removed is "transcript" field but not other fields. What is the idea of removing non-nullable value tests for other fields. Once these tests are removed, this test file only has tests for null field values, which doesn't seem to be what the change is for.

themes/base/templates/storyBlockTextEdit.handlebars
```diff
@@ -1,41 +1,25 @@
-<div lang="{{localizedMessages.language}}" class="sjrk-st-block">
+<fieldset lang="{{localizedMessages.language}}" class="sjrk-st-block">
+    <legend class="fl-hidden-accessible">{{localizedMessages.message_textBlock}}</legend>
     {{#getIds "storySelectionCheckbox"}}
         <div class="sjrk-st-label-container fl-hidden-accessible">
-            <label for="$id">{{../localizedMessages.message_blockSelectionCheckboxLabel}}</label>
```
cindyli, line 4: The translation of `message_blockSelectionCheckboxLabel` in `themes/base/messages/storyBlockMessages_es.json` and `themes/base/messages/storyBlockMessages_pt.json` should be replaced by labels for text, audio, image, video.

tests/ui/js/block-timeBasedTests.js
BlueSlug: The purpose of this test sequence is to test the model relay that combines the various values in the block into a string. There were previously four values being combined, so I was testing (and numbering) every possible combination, but since we've removed one of the values, half of the tests are no longer useful and so I've removed them. The remaining 8 test the various combinations of the other three values.

themes/base/templates/storyBlockTextEdit.handlebars
BlueSlug: Indeed. Since we disabled the localization functions, the message bundles for other languages have fallen out of date. This may take some time to complete

BlueSlug: Should I file a Jira and address this at a later time, or would you prefer that I make these updates now? I don't speak Portuguese, so that might just be a matter of adding the keys with empty strings (@gtirloni contributed the `pt` bundle).

cindyli: You said the localization has been disabled. It might make sense to have this work go into that JIRA.

tests/ui/js/block-timeBasedTests.js
cindyli: It would be helpful to organize these tests in a way that covers: 1. your purpose of the sequence such as from 111 -> 000; 2. test every value with firing request in null and non-null values. 

BTW, your comment of 4 digits, such as 1010, can now be changed to 3 digits, as now it only supports 3 values.

cindyli: Just find you've done the digit change. Thanks.

PR #50, opened 2019-11-08
SJRK-286: images rotate to correct orientation on save

BlueSlug:
This update adds four new node packages, does some refactoring of server code and adds a couple of new steps to the manual tests for the `imageBlockEditor`.


src/server/requestHandlers.js
```diff
@@ -1,431 +1,490 @@
...
+                });
+
+                rotateImagePromises.push(singleRotatePromise);
+            }
+        });
+
+        // using [""] notation for .catch() and .finally() to pass linting,
+        // at least until FLUID-6419 is addressed:
+        // https://issues.fluidproject.org/browse/FLUID-6419
+        Promise.all(rotateImagePromises)["catch"](function (error) {
```
cindyli: `rotateImagePromises` contains an array of infusion promises. Is there a reason that you decided to use JS standard `Promise.all()` instead of `[infusion fluid.promise.sequence()](https://docs.fluidproject.org/infusion/development/PromisesAPI.html#fluidpromisesequencesources-options)` to handle it? Using fluid.promise.sequence(), functions attached for `catch` and `finally` can be more proper `resolve` and `reject` callback functions.

src/server/requestHandlers.js
```diff
@@ -1,431 +1,490 @@
...
+});
+
+sjrk.storyTelling.server.handleSaveStoryWithBinaries = function (request, dataSource, authoringEnabled) {
+    if (authoringEnabled) {
+        var rotateImagePromises = [];
+
+        // rotate any images based on their EXIF data, if present
+        fluid.each(request.req.files.file, function (singleFile) {
+            if (singleFile.mimetype && singleFile.mimetype.indexOf("image") === 0) {
+                var singleRotatePromise = sjrk.storyTelling.server.rotateImageFromExif(singleFile).then(null, function (error) {
```
cindyli: If you attach `then()` here, the promise returned by `rotateImageFromExif` will be resolved here. It doesn't make sense to push `singleRotatePromise` into an array and use `Promise.all` or `fluid.promise.sequence` later to wait for their executions because at then promises in the array have already been resolved here. If you do expect to have one promise returned when all `singleRotatePromise` are executed, remove the `then()` resolver from here and attach resolve/reject functions on the promise returned by `fluid.promise.sequence`.

src/server/requestHandlers.js
```diff
@@ -1,431 +1,490 @@
...
+// Rotates an image to be oriented based on its EXIF orientation data, if present
+sjrk.storyTelling.server.rotateImageFromExif = function (file, options) {
+    var togo = fluid.promise();
+
+    try {
+        // ensure the file is present and we have all permissions
+        fse.accessSync(file.path);
+
+        // apply default rotation options if none are provided
+        options = options || { quality: 85 };
```
cindyli: Can you add a code comment to explain why the default rotation options is set to `{ quality: 85 }`? Thanks.

cindyli:
I hit this error at clicking "publish my story" button:
`Publishing failed: Promise.all(...).catch(...).finally is not a function`.

This is to publish a story containing one single image. This issue can be reproduced across different browsers.

Screen Shot 2019-11-11 at 2 45 07 PM

tests/server/testServerWithStorage.js
```diff
@@ -186,10 +218,12 @@ sjrk.storyTelling.server.testServerWithStorageDefs = [{
         "onBlankStoryWithEmptyMediaBlocksSaveSuccessful": null
     },
     testUploadOptions: {
-        testFile: "./tests/testData/logo_small_fluid_vertical.png",
+        testPNGFile: "./tests/testData/logo_small_fluid_vertical.png",
+        testImageWithCorrectOrientation: "./tests/testData/obliterationroom.jpg",
```
cindyli: It might be helpful to rename images to "correctOrientation.jpg" and "incorrectOrientation.jpg".

src/server/requestHandlers.js
BlueSlug: The idea was to wait until all the promises had finished resolving, but that the final state of them is not relevant, since some of the promises will reject with an "error" message even if everything is okay. E.g. if the image provided has no EXIF data and nothing needs to be done, or if the image is already at Orientation 1.

I suppose I could use `fluid.promise.sequence()` and then chain the `finally` and `catch` off that last one, but do `fluid` promises have those methods? I thought that wasn't the case.

src/server/requestHandlers.js
BlueSlug: It was an arbitrary value I chose which was of decently high quality without going for 100%, which would generate the largest files. Is that a reasonable thing to put in a comment?

The default for quality when I call `jo.rotate` is 100, which I suppose is also okay. In that case, I would simply remove this line altogether.

tests/server/testServerWithStorage.js
BlueSlug: Good point! The `obliterationroom.jpg` image was already present, so rather than include a separate case to cover a correctly-oriented image, I decided to re-use that. Since that file is used in other tests, I think I'll find another image that has correct orientation and then call that `correctOrientation.jpg`

src/server/requestHandlers.js
amb26: Best to fix up the promise source so that it doesn't reject when everything is fine.

BlueSlug: I agree with @amb26's suggestion, though complication is introduced by the fact that my call to `jo.rotate` within `rotateImageFromExif()` will return an error code and reject its promise even if the result is acceptable. Here's the documentation, in case you're curious about all the possible cases: https://github.com/johansatge/jpeg-autorotate#error-handling

With that in mind, would it be reasonable to have a check on the error codes returned and, in the specific cases where everything is okay, resolve the promise and keep on rolling?
i.e. at this point: https://github.com/BlueSlug/sjrk-story-telling/blob/b084f7180f84cf406b82e518134711dfe5702669/src/server/requestHandlers.js#L232

If I add this error-code checking and resolve in those specific cases that are acceptable, then I could use `fluid.promise.sequence` and remove the awkward syntax workaround and ES6 class.

amb26: Yes, I think that is the best approach

BlueSlug:
Ready for more review

src/server/requestHandlers.js
```diff
@@ -1,431 +1,491 @@
...
+            }
+        });
+
+        fluid.promise.sequence(rotateImagePromises).then(function () {
+            var storyModel = JSON.parse(request.req.body.model);
+            var binaryRenameMap = sjrk.storyTelling.server.buildBinaryRenameMap(storyModel.content, request.req.files.file);
+
+            sjrk.storyTelling.server.saveStoryToDatabase(dataSource, binaryRenameMap, storyModel, request.events.onSuccess, request.events.onError);
+        }, function (error) {
+            if (!error.errorCode) {
```
cindyli: Why does `request.events.onError` only get fired when `!error.errorCode`? As to line 232-236, `errorCode` could also present when Image rotation fails.

src/ui/uploader.js
```diff
@@ -64,15 +64,40 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         }
     });
 
-    /* Updates the uploader's internal file representation with
-     * information stored in the DOM's uploader element.
+    /* Updates the uploader's internal file representation with information
+     * stored in the DOM's uploader element. If the file is an image, an attempt
```
cindyli, line 68: The comment here says "an attempt will be made to rotate images", which doesn't seem to happen in this function. Do I miss anything?

src/ui/uploader.js
```diff
@@ -64,15 +64,40 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
      * - "that": the uploader itself
      * - "fileInput": the DOM uploader element
      */
     sjrk.storyTelling.block.singleFileUploader.handleFileInputChange = function (that, fileInput) {
         var fileList = fileInput[0].files;
         var currentFile = fileList[0];
-        that.currentFile = currentFile;
+
+        if (currentFile && currentFile.type.indexOf("image") === 0) {
+            loadImage(currentFile, function (img) {
```
cindyli, line 84: Can you add code comments about the purpose of using `loadImage()` and `.toBlob()`? I understand `loadImage()` is to extract the image's EXIF data but don't know what `.toBlob()` is for. Thanks.

tests/server/testServerWithStorage.js
```diff
@@ -520,9 +639,148 @@ sjrk.storyTelling.server.testServerWithStorageDefs.testImageRetrieval = function
...
+        if (filePath) {
+            var oldFilePath = "./tests/testData/" + testCase.fileName;
+            filePath = "./tests/server/uploads/" + uuidv1() + path.extname(testCase.fileName);
+            fs.copyFileSync(oldFilePath, filePath);
+        }
+
+        // call the function passing the new path and options where applicable
+        jqUnit.stop();
+        sjrk.storyTelling.server.rotateImageFromExif({ path: filePath }, testCase.options).then(function () {
+            jqUnit.assertEquals("Rotation call resolved as expected", testCase.expectedResolution, true);
```
cindyli: I think the tests for `rotateImageFromExif` should check images have been rotated to correct orientations. Why are assertions here and at line 719 checking resolutions rather than orientations?

src/ui/uploader.js
BlueSlug: `.toBlob()` is used because the call to `.loadImage()` returns a `canvas` element, but what we want is file data (a blob) that we can include in the form. We know it's a canvas that's being returned because setting the `orientation` option to `true` will also enable the `canvas` option in the call to `loadImage`: https://github.com/blueimp/JavaScript-Load-Image#options

BlueSlug: I'll add comments for sure!

src/ui/uploader.js
BlueSlug: indeed, the call to `loadImage` is the attempt to rotate images (as you suggested below, I'll be providing some more info on this)

tests/server/testServerWithStorage.js
BlueSlug: I have added assertions to the two cases where the image data is returned, i.e. the promise resolves but without "errors". The `correctOrientation` image will resolve but won't return image data since technically there was an error in the call to `jo.rotate`

BlueSlug:
Ready for more review

tests/server/testServerWithStorage.js
```diff
@@ -687,13 +687,13 @@ sjrk.storyTelling.server.testServerWithStorageDefs.rotateImageFromExifTests = fu
         { fileName: "", options: null, expectedResolution: false },
         { fileName: "", options: "", expectedResolution: false },
         { fileName: "correctOrientation.jpg", options: null, expectedResolution: true },
-        { fileName: "incorrectOrientation.jpeg", options: null, expectedResolution: true },
+        { fileName: "incorrectOrientation.jpeg", options: null, expectedResolution: true, expectedOrientation: 1 },
```
cindyli: Thanks for the improvement, @BlueSlug. It would be more helpful to further improve the test to show what processing results you expect from sjrk.storyTelling.server.rotateImageFromExif() when it deals with different media types. You have quite comprehensive test files with images, audio, video, but the test for them is mostly to show they are all resolved by this function. The test doesn't show what you expect this function does to them? For example, if you expect this function does nothing to audios, videos and images with correct orientations, add steps to show orientation values, or some other indicators, of these files stay unchanged before and after the rotation, while orientation values do change for images with incorrect orientations.

BlueSlug:
Ready for more review

cindyli:
Merged at c131d003a685a7bc29fdd9a8f787351d4ec9bd49

PR #51, opened 2019-11-13
SJRK-292: Remove extraneous comma from story browse page

greatislander:
As discussed yesterday.


BlueSlug (changes requested):
The colon was in fact moved to the message bundle in order to account for possible localizations that don't use a comma in this context, so it may be more appropriate to fix the template rather than the string.

BlueSlug:
I should have provided this context in the Jira, sorry!

greatislander:
Will adjust, @BlueSlug!

greatislander:
@BlueSlug Would this mean the colon should also be moved from template to messages for `Contains:`?

BlueSlug:
> @BlueSlug Would this mean the colon should also be moved from template to messages for `Contains:`?

Yes, @greatislander, that's a good catch!

greatislander:
@BlueSlug I've done this too in b888009.

BlueSlug:
Thanks @greatislander, everything looks good to me, build and tests pass, and I would be fine with this being merged in.

cindyli:
Shall a colon be added for [the Farsi translation of "keywords"](https://github.com/greatislander/sjrk-story-telling/blob/SJRK-292/themes/base/messages/storyMessages_fa.json#L31)?

cindyli:
Shall we take the chance to sync up this rule of defining colon in message bundle files instead of templates?

Searching the colon across .handlebars files shows colons still appear after these messages: message_browseViewText, message_languageLabelText, message_shareResponseText.

BlueSlug:
@cindyli that seems like a reasonable course of action

greatislander:
@BlueSlug @cindyli I updated the remaining messages as per @cindyli's comment. One thing I noticed is that the `message_browseViewText` is uppercased in English but titlecased in other localizations. I'm wondering if it might make sense to use CSS to uppercase this rather than having the string uppercased? (That appears to be how the `List` / `Grid` toggles are uppercased.)

cindyli:
>I'm wondering if it might make sense to use CSS to uppercase this rather than having the string uppercased?

Sounds a good idea to me. @BlueSlug, what do you think?

BlueSlug:
@greatislander @cindyli this is an interesting question. In the case of the list/grid view setting for the Browse page, I'm not actually sure that we should be applying the case via CSS, again because this may change from localization to localization. I can't think of a particular language/locale that we _wouldn't_ want to apply this for, but if we're being careful enough to isolate our colons then perhaps we should take similar steps with the casing...

I'm a little torn on this one! TL;DR I'm not sure, but I agree that the approach should be consistent across the site.

cindyli:
@BlueSlug brought up a good point that "the casing may change from localization to localization".

For this pull request, probably we should only deal with the uppercased "VIEW" and leave the casing to a broader discussion.

greatislander:
> … I'm not actually sure that we should be applying the case via CSS, again because this may change from localization to localization.

It appears that your instinct is correct, @BlueSlug — I found this on MDN which specifically address CSS case manipulation and localization: https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices#CSS_issues

So, it seems that it would also be a good idea to change Grid / List to GRID / LIST in the messages file and remove the CSS rule that uppercases them. Make sense?

BlueSlug:
> So, it seems that it would also be a good idea to change Grid / List to GRID / LIST in the messages file and remove the CSS rule that uppercases them. Make sense?

@greatislander yes, that seems like the best choice, especially if MDN is recommending it :)

greatislander:
Okay, I think we're done :)

cindyli:
Merged at ceba0103e38470c49c550c7b02affc563edd1175

PR #56, opened 2020-02-06
SJRK-104: Use pushState to manage back/forward behaviour when creating a story

BlueSlug:
This update implements the task as described in [the Jira](https://issues.fluidproject.org/browse/SJRK-104). I made the choice not to include the state variables in the query string, but this is easily enabled by removing the exclusion rules in the `historian` component.

This update also addresses [SJRK-261](https://issues.fluidproject.org/browse/SJRK-261).


BlueSlug:
Please hold off reviewing for the time being, I will be adding tests specific to the `historian` component and solving some sequence breaks that existed before this work

BlueSlug:
Ready for review.

I tried to add tests for the forward & back functionality, but I was getting browser security related errors ("TypeError: Illegal invocation") when trying to trigger those actions programatically.

src/ui/base-page-storyEdit.js
```diff
@@ -62,6 +93,14 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             setEditorDisplay: {
                 funcName: "sjrk.storyTelling.base.page.storyEdit.setEditorDisplay",
                 args: ["{that}.options.selectors.mainContainer", "{that}.options.selectors.pageContainer", "{that}.options.pageSetup.authoringEnabled", "{that}.options.pageSetup.hiddenEditorClass"]
+            },
+            showEditorHidePreviewer: {
```
cindyli, line 129: `showEditorHidePreviewer()` and `hideEditorShowPreviewer()` invokers can be combined into one by accepting a boolean argument.

src/ui/ui-storyEditor.js
```diff
@@ -14,14 +14,33 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            "editStoryStepVisible": [{
+                this: "{that}.dom.storyEditStoryStep",
+                method: "toggle",
+                args: ["{change}.value"],
+                namespace: "manageEditStoryStepVisibility"
+            },
+            {
+                this: "{that}.dom.storyMetadataStep",
+                method: "toggle",
+                args: ["{change}.oldValue"],
```
cindyli: I understand you use `{change}.oldValue` to retrieve the opposite of the current boolean value in `{change}.value`. However, it introduces a logic confusion that this action is controlled by the old model value rather than the current. Personally, I prefer to sacrifice a little coding convenience for clarifying the logic.

src/ui/ui-storyEditor.js
```diff
@@ -96,25 +115,27 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-                    ["{that}.dom.storyEditorPage2"],
-                    ["{that}.dom.storyEditorPage1"],
-                    "{that}.events.onVisibilityChanged"
-                ]
-            },
             "onRemoveBlocksRequested.removeSelectedBlocks": {
                 funcName: "sjrk.storyTelling.ui.storyEditor.removeSelectedBlocks",
                 args: ["{that}", "{that}.blockManager.managedViewComponentRegistry"]
+            },
+            "onEditorNextRequested": {
```
cindyli: For consistency, include the namespace with the event as above. 

The same comment applies to the listener at line 126.

cindyli:
`editorVisible` and `editStoryStepVisible` work together to control 3 steps of editing and previewing. It would be nice to add code comment or documentation to explain model values of these 2 at different steps.

src/ui/base-page-storyEdit.js
```diff
@@ -16,6 +16,40 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         pageSetup: {
             hiddenEditorClass: "hidden"
         },
+        model: {
+            // the initial state is only the editor ui showing.
+            // in much the same way as within the editor, the visibility
+            // of the editor and previewer are mutually exclusive and
+            // represented by a single model value, for the time being
+            editorVisible: true
```
amb26: As below, with the use of "oldValue" which @cindyli has noted below. Rather than applying this historyful gimmick, you should have a properly organised model with separate flags for "editorVisible" and "previewerVisible". Connect them with a relay - presumably this is the situation which gave rise to the legendary standup update "Infusion doesn't have an inversion relay" : P
Looking at the impl for the fluid.transforms.linearScale transform which could have been hacked into service, I see that @kaspermarkus unhelpfully put some type checks into it, so I suggest you just apply
````
    sjrk.storyTelling.not = function (value) {
          return !value;
     };
````
and specify this as the relay rule.

Comment (amb26) [reply src/ui/base-page-storyEdit.js]: For extra helpfulness, please contribute this to Infusion as fluid.transforms.not together with its self-inverse definition and a couple of small test cases :)

File: src/ui/ui-storyEditor.js (hunk 2 above)
Comment (BlueSlug) [reply src/ui/ui-storyEditor.js]: I agree that this method is confusing. What I'd like is to have a way to say "the opposite of `{change}.value`", but without the `negate` transform I was beginning to create (my work on that can be found in [my WIP branch](https://github.com/BlueSlug/sjrk-story-telling/blob/fd1f8096dcbf1ddd98a484c7dacc130412d898bb/src/ui/transforms.js#L111-L126)), I'm not sure how best to approach this.

What do you think of adding a basic "not" function locally in this project, with the aim of using it in an expander and passing in `{change}.value`? This function can later be supplanted by the `negate` transform, whenever that ends up being incorporated into Infusion.

Comment (BlueSlug) [reply src/ui/ui-storyEditor.js]: I've just noticed @amb26's comment below suggesting to do something very similar

File: src/ui/base-page-storyEdit.js (hunk 4 above)
Comment (BlueSlug) [reply src/ui/base-page-storyEdit.js]: Indeed, this is precisely the situation that gave rise to my comments ;) I began to work on such a transform, but we decided that for the sake of expedience I would shelve that work and Just Get It Working™ with this binary view-control method, and come back to this when there is less pressing work to be done.

I've achieved a less confusing solution via [the method I proposed a couple review items above](https://github.com/fluid-project/sjrk-story-telling/pull/56#discussion_r377315295). It's a bit of a middle ground between using `{change}.oldValue` and re-implementing the model relay I once had (with either a free transform calling this new `not` function or as a full-fledged inversion transform). Not ideal, but perhaps good enough for now? I can file a Jira to make sure we repay this technical debt.

**Comment by BlueSlug** [comment]
After discussing the one value / two value situation with @cindyli and @amb26, we have two possible solutions to the matter at hand, each with its pros and cons:

1. we continue using a single model value to represent the visibility states (being the steps of the editor or the `storyEditor` & `storyPreviewer` within the Edit page), and give it a better name to more clearly convey that it controls more than one thing
2. we revert to using multiple model values to control each piece individually, and we add model relays to coordinate them

Approach 1 sacrifices the granularity of individual control values for simplicity of implementation. We don't currently have a need for more than two states per value, so the only programming change necessary with this approach is to choose a better name. When requirements change, the single value can be adapted to use string values instead of boolean, in much the same way as the model relay and listeners in `storyPreviewer`. Downsides include tigher coupling and a steeper learning curve for any new developers.

Approach 2 has the benefit of clarity from one model value representing one thing (i.e. the visibility of one discrete element of the UI) as well as more loosely coupled code to make future development less complicated. The downside is, of course, that there will need to be more wiring (model relays, listeners and transforms) to maintain the relationships between the various values in order to change them with one mutator call, or more mutator calls to change all the relevant values. Reintroducing those bits of wiring would take more time, and they may be more repetitive.

When asked in the chat a couple of days ago why approach 2 would be better, @amb26 said:

> because the purpose of model state is to be a mirror for visible state
Just because you think you have certain exclusion relations in place which preclude certain combinations of such state, there's no reason use these to economise on your addressible model space ....
After all, the relations might go away some day, as a result of some further authorial decision
And then there'll you'll be, stuck with a configuration of your application which is impossible to encode
Or.... there someone will be

So it would seem that approach 2 is superior in the long term, at the cost of further development time and a bit of premature optimization. The question I have at this point to those in the reviewer role is: which approach should I take in order to get these changes merged into `master`?

**Comment by BlueSlug** [comment]
Ready for review

File: tests/ui/js/transformsTests.js [hunk 5]
```diff
@@ -279,4 +279,139 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            expectedResult: false,
+            expectedInverseResult: true
+        },
+        "test_18": {
+            input: null,
+            expectedResult: true,
+            expectedInverseResult: false
+        },
+        "test_19": {
+            // this case does not resolve as expected. normally, !undefined === true
```
cindyli: If the result of this test case is not expected, shall the transform be fixed to handle it?

tests/ui/js/transformsTests.js
```diff
@@ -279,4 +279,139 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            expectedInverseResult: true
+        },
+        "test_23": {
+            input: -Infinity,
+            expectedResult: false,
+            expectedInverseResult: true
+        }
+    };
+
+    jqUnit.test("Test negate transform function", function () {
```
cindyli: Inverse results are defined but not tested.

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -406,22 +406,23 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-                        ["{storyEdit}.storyEditor.dom.storyEditorPage2", "{storyEdit}.storyPreviewer.container"],
-                        ["{storyEdit}.storyEditor.dom.storyEditorPage1"]
+                        ["{storyEdit}.storyEditor.dom.storyMetadataStep", "{storyEdit}.storyPreviewer.container"],
+                        ["{storyEdit}.storyEditor.dom.storyEditStoryStep"]
                     ]
                 },
                 {
-                    "jQueryTrigger": "click",
-                    "element": "{storyEdit}.storyEditor.dom.storyEditorNext"
+                    jQueryTrigger: "click",
```
cindyli, line 434: At the change from editor to previewer and vice versa, as well as from editor step to metadata step and vice versa, please check what expected to be tracked by gpii.locationBar are tracked correctly.

BlueSlug:
Ready for more review

tests/ui/js/transformsTests.js
```diff
@@ -253,32 +228,159 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                };
 
-            var resultString = fluid.model.transformWithRules(
-                { sourceArray: testCase.input },
-                { resultString: transformRules }
-            ).resultString;
-            jqUnit.assertEquals("Generated string value for test case " + index + " is as expected", testCase.expectedResult, resultString);
+                // test the transformation
+                var result = fluid.model.transformWithRules(testCase, transformRules);
+                var assertionFunction = typeof testCase.expectedResult === "object" ? jqUnit.assertDeepEq : jqUnit.assertDeepEq;
```
jobara: Looks like you have a typo because both sides of the conditional point at `jqUnit.assertDeepEq`. However, the tests are passing because [`jqUnit.assertDeepEq`](https://github.com/fluid-project/infusion/blob/c3c22dc4c3a3683280989534bb5f03f8d90d1bf9/tests/test-core/jqUnit/js/jqUnit.js#L173-L179) will drop down to `jqUnit.assertEquals` if the expected or actual values are primitives. With that in mind, maybe you just call `jqUnit.assertDeepEq` always and skip the conditional check completely.

BlueSlug: Nice catch! Not sure how that one slipped by. Thanks @jobara!

BlueSlug:
Ready for more review

src/ui/base-page-storyEdit.js
```diff
@@ -18,6 +18,58 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+        },
+        modelListeners: {
+            "editorVisible": [{
+                this: "{storyEditor}.container",
+                method: "toggle",
+                args: ["{change}.value"],
+                namespace: "manageEditorVisibility"
+            },
+            {
+                func: "{that}.events.onContextChangeRequested.fire",
```
cindyli, line 65: As the change of `model.editorVisible` automatically triggers the change of `model.previewerVisible` and vice versa, which means the change to any of these 2 model values will trigger `onContextChangeRequested` event to be fired twice. Is this correct?

src/ui/base-page-storyEdit.js
```diff
@@ -18,6 +18,58 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
         pageSetup: {
             hiddenEditorClass: "hidden"
         },
+        model: {
+            // the initial state is only the editor ui showing.
```
cindyli: Could you improve this comment to better explain 3 steps for editing a story and how the visibility of them are controlled. Thanks.

src/ui/base-page-storyEdit.js
```diff
@@ -77,17 +132,22 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             "video": ["mediaUrl"]
         },
         components: {
+            // manaages browser history for in-page forward-back support
+            historian: {
+                type: "gpii.locationBar",
+                options: {
+                    model: {
+                        editorVisible: "{storyEdit}.model.editorVisible",
```
cindyli, line 153: Add a code comment to explain that only these 2 model values are tracked because transformations between:
1. `{storyEdit}.model.editorVisible` and `{storyEdit}.model.previewerVisible`
2. `{storyEditor}.model.editStoryStepVisible` and `storyEditor}.model.metadataStepVisible`
will set other model values properly.

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -408,22 +424,23 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                    element: "{storyEdit}.storyEditor.dom.storyEditorNext"
                 },
                 {
-                    "event": "{storyEdit}.storyEditor.events.onVisibilityChanged",
-                    "listener": "sjrk.storyTelling.testUtils.verifyPageVisibility",
-                    "args": [
-                        ["{storyEdit}.storyEditor.dom.storyEditorPage1", "{storyEdit}.storyPreviewer.container"],
-                        ["{storyEdit}.storyEditor.dom.storyEditorPage2"]
+                    changeEvent: "{storyEdit}.storyEditor.applier.modelChanged",
```
cindyli, line 438: Add a visibility test for clicking the "previous" button from the metadata step to the edit story step.

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -747,6 +784,78 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                {
+                    "event": "{storyEdit}.events.onContextChangeRequested",
+                    listener: "sjrk.storyTelling.testUtils.verifyStepVisibility",
+                    args: [
+                        ["{storyEdit}.storyEditor.dom.storyEditStoryStep", "{storyEdit}.storyPreviewer.container"],
+                        ["{storyEdit}.storyEditor.dom.storyMetadataStep"]
+                    ]
+                },
+                {
+                    "jQueryTrigger": "click",
```
cindyli, line 802: Please ignore my comment above. You've got the test for it here. Sorry.

src/ui/base-page-storyEdit.js
BlueSlug: The double-firing is unintentional. I will remove the listener on `previewerVisible`

BlueSlug:
Ready for review!

src/ui/base-page-storyEdit.js
cindyli: bump

BlueSlug: I misread your suggestion, have added a comment here :)

cindyli:
Merged at 94a55d18c52e605dc9c7292fd0bc75b6d8887b03

PR #57, opened 2020-02-13
SJRK-267: Move to a centralized copyright notice

BlueSlug:
This change addresses [SJRK-267](https://issues.fluidproject.org/browse/SJRK-267) and moves all copyright dates and contributor listings into a central file (AUTHORS.md). This is consistent with the approach taken in Infusion.

At a minimum, the annual updating of the copyright dates now only has to be done in one file, rather than 80.


LICENSE.txt
```diff
@@ -1,12 +1,15 @@
 The Storytelling Tool is available under the terms of the New BSD license.
 As a recipient of Storytelling Tool, you may choose which license to receive this
 code under (except as noted in per-module LICENSE files). All modules are
-Copyright 2017-2020 OCAD University except where noted otherwise in the code itself,
-or if the modules reside in a separate directory, they may contain explicit
-declarations of copyright in both the LICENSE file in the directory in which they
-reside and in the code itself. No external contributions are allowed under
-licenses which are fundamentally incompatible with the BSD license that
-the Storytelling Tool is distributed under.
+Copyright The Storytelling Tool copyright holders except where noted otherwise in
```
cindyli: Please fix this sentence of "All modules are Copyright The Storytelling Tool copyright holders...".

BlueSlug: That's a fair point! This is the same language that the Infusion notices use, though I agree it's clunky and potentially confusing.

As we discussed in person, I will simply remove that line altogether and more explicitly direct users to AUTHORS.md for copyright info. I'm also tempted to remove the parts about which licence it uses, since that is also covered in AUTHORS.md, README.md and LICENSE.txt.

The copyright is assigned automatically at authoring time, so legally speaking removing that line doesn't open us up to any risk. The license I'm less sure about so I'll leave that alone.

cindyli:
Merged at 7a35d8dc297e1d39567aa1751182326337cc7b07

PR #59, opened 2020-03-17
SJRK-326: solved page state bug caused by saving page model in cookie

BlueSlug:
The bug was due to the entire `page` component's model being saved to the `cookieStore`, including the view state values.

The solution is to keep the stored values in a specific collection called `persistedValues`. The `uiLanguage` and `storyBrowseDisplayPreference` keys have been added to that collection, as nothing else at this point needs to be persisted, and tests have been adjusted accordingly.

### DIFF
--- a/src/ui/base-page-storyBrowse.js
+++ b/src/ui/base-page-storyBrowse.js
@@ -17,7 +17,9 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     fluid.defaults("sjrk.storyTelling.base.page.storyBrowse", {
         gradeNames: ["sjrk.storyTelling.base.page"],
         model: {
-            storyBrowseDisplayPreference: ""
+            persistedValues: {
+                storyBrowseDisplayPreference: ""
+            }
         },
         events: {
             onAllUiComponentsReady: {
@@ -35,7 +37,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
                     listeners: {
                         "onViewChangeRequested.savePreference": {
                             func: "{storyBrowse}.applier.change",
-                            args: ["storyBrowseDisplayPreference", "{arguments}.0.data"],
+                            args: [["persistedValues", "storyBrowseDisplayPreference"], "{arguments}.0.data"],
                             priority: "first"
                         }
                     },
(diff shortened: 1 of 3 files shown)


src/ui/base-page-storyBrowse.js
```diff
@@ -17,7 +17,9 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     fluid.defaults("sjrk.storyTelling.base.page.storyBrowse", {
         gradeNames: ["sjrk.storyTelling.base.page"],
         model: {
-            storyBrowseDisplayPreference: ""
+            persistedValues: {
```
cindyli, line 20: Could you add a code comment to explain:
1. what will be tracked in the model;
2. what will be tracked in `model.persistedValues`;
3. the goal of having `model.persistedValues`.

BlueSlug: I've left a comment in `base-page.js` which outlines the use of that collection, but I'll add these clarifications

BlueSlug: What do you think of the following, if it's on the `base-page` grade?
```
// only values in this colleciton will be persisted by the cookieStore
// the goal of separating them is to allow the use of other model values
// such as the view state values in `sjrk.storyTelling.base.page.storyEdit`
// without saving them.
```

cindyli: Yes, it's ok to have the comment on the `base-page` grade. It would be better if you could also write about what will be tracked in this collection for persistence. Thanks.

BlueSlug: Okay, here's a revised version of that comment:

```
// Only values in this colleciton will be persisted by the cookieStore.
//
// Currently, those values are the `uiLanguage` in this grade and
// `storyBrowseDisplayPreference` in `sjrk.storyTelling.base.page.storyBrowse`.
//
// The goal of separating them is to allow the use of other model values
// such as the view state values in `sjrk.storyTelling.base.page.storyEdit`
// without saving them.
```

cindyli: Looks great. Thanks.

BlueSlug:
Ready for further review :)

cindyli:
Merged at b886c18b792ea90b4a85cfc8b0e8a7028ea11361

PR #71, opened 2020-04-17
SJRK-289: Add auto-save or caching feature for story authoring

BlueSlug:
This update adds story autosaving in the Edit page, and involves a small amount of refactoring of the `ui` grades (as detailed in the commit comment for 0481597).

**PLEASE NOTE**
The work is still in progress, and I've created this pull request in order to get review started as soon as possible ahead of release.

Still remaining:
- Loading of media files with their respective blocks (saving is already implemented)
- Tests for loading blocks
- Tests for saving and loading media files


cindyli:
The restore of the auto save only restores contents for text blocks but not other blocks of image, audio and video. 

This error is reported at restoring media contents which might be related to why they cannot be restored: `Security Error: Content at http://localhost:8081/storyEdit.html may not load data from blob:http://localhost:8081/0a42283f-5e65-e244-b4ec-2d916be501a4.`.

BlueSlug:
> The definition of `blockTypeLookup` here doesn't seem to be needed as it has been defined in [the new `storyUi` component](https://github.com/fluid-project/sjrk-story-telling/blob/0481597ccdb791f283004cac0efeb9e8531b3803/src/ui/ui.js#L142) that `storyEditor` will inherit from.

This is actually an override with the `blockUi.editor` grades for each block type, whereas the grades listed in `storyUi` are viewer grades.

cindyli:
> > The definition of `blockTypeLookup` here doesn't seem to be needed as it has been defined in [the new `storyUi` component](https://github.com/fluid-project/sjrk-story-telling/blob/0481597ccdb791f283004cac0efeb9e8531b3803/src/ui/ui.js#L142) that `storyEditor` will inherit from.
> 
> This is actually an override with the `blockUi.editor` grades for each block type, whereas the grades listed in `storyUi` are viewer grades.

I see. Thanks.

BlueSlug:
> `onBlockManagerCreated` doesn't have listeners attached. Can it be removed?

This is used in the `storyEditor` grade as part of its compound event `onReadyToBind`

cindyli:
> > `onBlockManagerCreated` doesn't have listeners attached. Can it be removed?
> 
> This is used in the `storyEditor` grade as part of its compound event `onReadyToBind`

Right. Missed that. Thanks.

cindyli:
> > The definition of `blockTypeLookup` here doesn't seem to be needed as it has been defined in [the new `storyUi` component](https://github.com/fluid-project/sjrk-story-telling/blob/0481597ccdb791f283004cac0efeb9e8531b3803/src/ui/ui.js#L142) that `storyEditor` will inherit from.
> 
> This is actually an override with the `blockUi.editor` grades for each block type, whereas the grades listed in `storyUi` are viewer grades.

If `editor` and `viewer` have their own unique block types, it would be less confusing to move `blockTypeLookup` out of the shared `storyUI` into individual components.

BlueSlug:
> > > The definition of `blockTypeLookup` here doesn't seem to be needed as it has been defined in [the new `storyUi` component](https://github.com/fluid-project/sjrk-story-telling/blob/0481597ccdb791f283004cac0efeb9e8531b3803/src/ui/ui.js#L142) that `storyEditor` will inherit from.
> > 
> > 
> > This is actually an override with the `blockUi.editor` grades for each block type, whereas the grades listed in `storyUi` are viewer grades.
> 
> If `editor` and `viewer` have their own unique block types, it would be less confusing to move `blockTypeLookup` out of the shared `storyUI` into individual components.

I'd considered this, since currently the `storyViewer` is the only grade making use of the values defined in `blockManager`, but I left it as you see it because that collection is passed in as a function argument in the invoker `createBlocksFromData`. I'll have to think about how, exactly, to untangle that part.

cindyli:
> > > > The definition of `blockTypeLookup` here doesn't seem to be needed as it has been defined in [the new `storyUi` component](https://github.com/fluid-project/sjrk-story-telling/blob/0481597ccdb791f283004cac0efeb9e8531b3803/src/ui/ui.js#L142) that `storyEditor` will inherit from.
> > > 
> > > 
> > > This is actually an override with the `blockUi.editor` grades for each block type, whereas the grades listed in `storyUi` are viewer grades.
> > 
> > 
> > If `editor` and `viewer` have their own unique block types, it would be less confusing to move `blockTypeLookup` out of the shared `storyUI` into individual components.
> 
> I'd considered this, since currently the `storyViewer` is the only grade making use of the values defined in `blockManager`, but I left it as you see it because that collection is passed in as a function argument in the invoker `createBlocksFromData`. I'll have to think about how, exactly, to untangle that part.

One solution is to leave `blockTypeLookup: null` in the `ui` component, along with a code comment to mention this lookup map must be provided by integrator components.

BlueSlug:
As discussed, I have removed the file saving/loading code and replaced it with a function that clears out any imageUrl or mediaUrl values on load.

Ready for more review

src/ui/base-page-storyEdit.js
```diff
@@ -208,6 +251,89 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+     * @param {String} sourceName - the name of the Infusion change source for this update
+     */
+    sjrk.storyTelling.base.page.storyEdit.loadStoryFromAutosave = function (storyAutosaveKey, storyEditor, storyPath, sourceName) {
+        try {
+            // localStorage can only store string values
+            var savedStory = JSON.parse(window.localStorage.getItem(storyAutosaveKey));
+
+            if (savedStory) {
+                // media block blob URLs are no longer valid/reliable after a page reload
+                sjrk.storyTelling.base.page.storyEdit.clearMediaBlockUrls(savedStory.content);
```
cindyli, line 291: I think it's better to clear media block urls before the save rather than at reading back the saved.

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -862,6 +918,113 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                    args: ["/stories/", "", "application/json"]
+                },
+                {
+                    "jQueryTrigger": "click",
+                    "element": "{storyEdit}.storyPreviewer.dom.storyShare"
+                },
+                {
+                    funcName: "sjrk.storyTelling.testUtils.teardownMockServer"
+                },
+                // ensures the story was deleted upon successful publishing
```
cindyli: "the story" -> "the saved story"?

src/ui/base-page-storyEdit.js
BlueSlug: If I clear the block URL's on saving, then they will be nulled out while the user is still editing which will cause unexpected behaviour (e.g. images will look right but not publish). I think it's better to do it on load, since we know that's a recovery situation and it's okay to clear them.

cindyli: A reasonable point and I agree.

PR #75, opened 2020-04-29
NOJIRA: CSS fixes

BlueSlug:
This should be merged into `stories-aihec-production` after it is merged into `master`

### DIFF
--- a/themes/learningReflections/css/learningReflections.css
+++ b/themes/learningReflections/css/learningReflections.css
@@ -224,7 +224,8 @@ h3:empty {
 
 .sjrk-st-lr-knockout {
     background: var(--site-orange);
-    color: #fff
+    color: #fff;
+    padding-top: 2rem !important;
 }
 
 .sjrk-st-intro-collection {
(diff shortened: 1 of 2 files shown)


themes/learningReflections/css/learningReflections.css
```diff
@@ -224,7 +224,8 @@ h3:empty {
 
 .sjrk-st-lr-knockout {
     background: var(--site-orange);
-    color: #fff
+    color: #fff;
+    padding-top: 2rem !important;
```
cindyli: Using `!important` is not a good css practice. Probably consider to use a css selector that has higher specificity such as `section.sjrk-st-lr-knockout`.

BlueSlug: I agree that it's good to avoid using `!important` when possible, since it can clutter up styling rules. I'll see what I can come up with :)

BlueSlug: Sorry I just noticed you suggested a higher-specificity selector! Using that

BlueSlug: I also targeted `.sjrk-st-lr-knockout section`, as the Introduction page uses the `knockout` class on an element that isn't a `section`

PR #77, opened 2020-05-14
SJRK-289: Add server-side auto-save to include files

BlueSlug:
This work implements server-side auto-saving of the story during its creation. This required totally reworking the way stories are created and published in the first place: rather than saving to database at publish time, the story is saved when the Edit page is first loaded, then when the user hits "publish", it switches a flag to indicate the story is ready for public consumption.

This work also encapsulates the following issues to varying degrees:
- [SJRK-185](https://issues.fluidproject.org/browse/SJRK-185) is dealt with, using UUID v4 now
- [SJRK-175](https://issues.fluidproject.org/browse/SJRK-175) is partly done, as the `videoBlockEditor` grade is now a `mediaBlockEditor`. Much remains to be done on this Jira, to be handled on another occasion


src/server/requestHandlers.js
```diff
@@ -105,36 +105,28 @@ fluid.defaults("sjrk.storyTelling.server.getStoryHandler", {
...
-        });
-
-        var responseAsJSON = JSON.stringify(response);
-        request.events.onSuccess.fire(responseAsJSON);
+            var responseAsJSON = JSON.stringify(response);
+            request.events.onSuccess.fire(responseAsJSON);
+        } else {
+            request.events.onError.fire({
+                isError: true,
+                message: "An error occurred while retrieving the requested story"
```
cindyli: It would be helpful to provide a clearer error message such as "Unauthorized: cannot access an unpublished story". If you'd rather not to expose the stories status to the end user, perhaps just return the unauthorized error but track the actual reason of being unauthorized in the server log.

src/server/requestHandlers.js
```diff
@@ -199,20 +166,24 @@ sjrk.storyTelling.server.handleSaveStoryWithBinaries = function (request, dataSo
...
  * @param {Component} dataSource - an instance of sjrk.storyTelling.server.dataSource.couch.story
- * @param {Object.<String, String>} binaryRenameMap - a map of uploaded file names to paths
  * @param {Object} storyModel - a single story's model
  * @param {Object} successEvent - an infusion event to fire upon successful completion
  * @param {Object} failureEvent - an infusion event to fire on failure
  */
-sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenameMap, storyModel, successEvent, failureEvent) {
-    var id = uuidv1();
+sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, storyModel, successEvent, failureEvent) {
+    var id = "";
```
cindyli: line 178-184 can be simplified to: 
```
var id = storyModel && storyModel.id ? storyModel.id : uuidv4();
```

src/server/requestHandlers.js
```diff
@@ -222,45 +193,86 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
+            });
+        } catch (ex) {
+            request.events.onError.fire({
+                isError: true,
+                message: "An error occured while saving the file:" + ex
+            });
         }
-    });
+    } else {
+        // delete the file?
```
cindyli: Can this comment be removed as the file hasn't been saved?

src/server/requestHandlers.js
```diff
@@ -222,45 +193,86 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
+        handleRequest: {
+            funcName: "sjrk.storyTelling.server.handleSaveStoryFile",
+            args: ["{arguments}.0", "{server}.storyDataSource", "{server}.options.globalConfig.authoringEnabled"]
+        }
+    }
+});
+
+/**
+ * Saves a single file to the server filesystem and responds to the HTTP request.
+ * If authoring is not enabled, the request raises an error.
```
cindyli: In this comment, please cover all cases that an error will be raised.

src/server/requestHandlers.js
```diff
@@ -222,45 +193,86 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
 
-/**
- * Update any media URLs to refer to the changed file names
- *
- * @param {Object} blocks - a collection of story blocks with file references
- * @param {Object} files - a list of files to refer to
+// Kettle request handler for saving a single file associated with a pre-existing story
+fluid.defaults("sjrk.storyTelling.server.saveStoryFileHandler", {
+    gradeNames: "kettle.request.http",
+    // a DataSource to save a single story along with any files it has
```
cindyli: This comment is inaccurate. Probably remove it as the comment at line 196 is more accurate.

src/server/requestHandlers.js
```diff
@@ -222,45 +193,86 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
+        }
+    }
+});
+
+/**
+ * Saves a single file to the server filesystem and responds to the HTTP request.
+ * If authoring is not enabled, the request raises an error.
  *
- * @return {Object.<String, String>} - a key-value collection linking uploaded filenames to server paths
+ * @param {Object} request - a Kettle request with data and files associated with a single story
```
cindyli: This request only contains one single file. Please update. Thanks.

src/ui/base-page-storyEdit.js
```diff
@@ -149,6 +142,51 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             clearAutosave: {
                 funcName: "sjrk.storyTelling.base.page.storyEdit.clearAutosave",
                 args: ["{that}.options.pageSetup.storyAutosaveKey"]
+            },
+            initializeStory: {
```
cindyli, line 200: Please add code comments to explain how this and following functions are used at different stages of the auto save and publish.

src/server/requestHandlers.js
```diff
@@ -222,45 +193,86 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
-
-                // If we find a match, update the media URL. If not, clear it.
-                if (mediaFile) {
-                    sjrk.storyTelling.server.setMediaBlockUrl(block, mediaFile.filename);
-                    binaryRenameMap[mediaFile.originalname] = mediaFile.filename;
-                } else {
-                    sjrk.storyTelling.server.setMediaBlockUrl(block, null);
+sjrk.storyTelling.server.handleSaveStoryFile = function (request, dataSource, authoringEnabled) {
+    if (authoringEnabled) {
+        try {
```
cindyli: This `try/catch` is not quite necessary as the reject callback of `getStoryPromise` will catch most failure. Once `id` at line 225 is checked, the `try/catch` can be removed and this deeply nested block will have one level less. To do so, line 222-226 can be refactored to:

```
var id = request.req.params.id;
if (authoringEnabled && id) {
    var getStoryPromise = ...
} else {
    // report error
};
```

src/server/middleware/saveStoryFile.js
```diff
@@ -55,8 +48,8 @@ fluid.defaults("sjrk.storyTelling.server.middleware.saveStoryWithBinaries", {
  * @param {Object} file - an uploaded file to process
  * @param {filenameResolverCallback} cb - the callback
  */
-sjrk.storyTelling.server.middleware.saveStoryWithBinaries.filenameResolver = function (req, file, cb) {
-    var id = uuidv1();
+sjrk.storyTelling.server.middleware.saveStoryFile.filenameResolver = function (req, file, cb) {
+    var id = uuidv4();
```
cindyli: A problem with generating a new id for every uploaded file is, if a user operates on a same media block by changing the file a few times, all these files will be kept on the server even though the last one is the one that should be kept. Those previous uploads won't be able to be removed even by deleting the story because the story only has the url of the last upload for that media block.

One solution is to generate this id and compose the file name at the client side. The server removes files with the same id before saving the new file. The removal is needed because the file extension can be different.

src/server/requestHandlers.js
BlueSlug: I've been going back and forth on this one, not sure whether I should be worried about enumeration attacks. Since we're using v4 UUID's, it's not impossible to guess ID's, but it's going to be very difficult and not a very practical vector, IMO.

If I'd like to avoid giving away that a story ID was correct but that it's unpublished, then it would be better to match this error message to the one below it. I'll make them both the same and send the actual error message on the server log, in each case.

src/server/requestHandlers.js
BlueSlug: Good catch! I'd left it that way to easily set breakpoints, but that's no longer necessary

src/ui/base-page-storyEdit.js
```diff
@@ -274,50 +312,70 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+    sjrk.storyTelling.base.page.storyEdit.saveNewStoryToServer = function (storySaveUrl, story, storyIdPath, sourceName, errorEvent) {
+        var serverSavePromise = sjrk.storyTelling.base.page.storyEdit.saveStoryToServer(storySaveUrl, story.model);
+
+        serverSavePromise.then(function (data) {
+            var successResponse = JSON.parse(data);
+
+            // store the ID on the story model for later use
+            story.applier.change(storyIdPath, successResponse.id, null, sourceName);
+        }, function (jqXHR, textStatus, errorThrown) {
+            fluid.log(fluid.logLevel.WARN, "Something went wrong");
```
cindyli: Be more specific with "Something went wrong". Replace it with a clearer error such as "Error at saving a new story to server."

src/server/requestHandlers.js
BlueSlug: The file is actually saved before this requestHandler is ever called, so it does in fact persist in the server file system even if authoring is disabled.

src/ui/uploader.js
```diff
@@ -129,11 +155,11 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
                 type: currentFile.type
             };
 
-            URL.revokeObjectURL(that.model.fileObjectURL);
-            that.applier.change("fileObjectURL", URL.createObjectURL(currentFile));
+            URL.revokeObjectURL(that.model.fileObjectUrl);
+            that.applier.change("fileObjectUrl", URL.createObjectURL(currentFile));
             that.applier.change("fileDetails", fileDetails);
```
cindyli: Is `fileDetails` saved for the testing purpose? It doesn't seem to be used in the implementation code.

tests/server/testServerWithStorage.js
```diff
@@ -611,6 +761,31 @@ sjrk.storyTelling.server.testServerWithStorageDefs.verifyStoryPersistence = func
...
+ *
+ * @param {String} data - the data returned by the call
+ * @param {Object} completionEvent - an event to fire on test completion
+ * @param {Boolean} authoringEnabled - a flag indicating whether authoring is enabled
+ */
+sjrk.storyTelling.server.testServerWithStorageDefs.verifyStoryGetFails = function (data, completionEvent, authoringEnabled) {
+    var parsedData = JSON.parse(data);
+
+    if (authoringEnabled) {
+        jqUnit.assertEquals("Story request produced an error as expected", true, parsedData.isError);
```
cindyli: Please verify error message, at least partial message, to make sure the rejection is due to the unpublished state.

tests/server/testServerWithStorage.js
```diff
@@ -611,6 +761,31 @@ sjrk.storyTelling.server.testServerWithStorageDefs.verifyStoryPersistence = func
...
+
+        if (completionEvent) {
+            completionEvent.fire(parsedData);
+        }
+    } else {
+        jqUnit.assert("Saved story data does not exist");
```
cindyli, line 751: Shall here verify an error message if `authoringEnabled === false` rather than a fake `assert`?

tests/server/testServerWithStorage.js
```diff
@@ -327,23 +376,49 @@ sjrk.storyTelling.server.testServerWithStorageDefs = [{
     },
     // Story with an image
     {
+        // initializes the story in its unpublished form
```
cindyli, line 364: Line 373-616 is a very long sequence. It should be broken down to a few sequences, each testing a meaning feature.

tests/server/testServerWithStorage.js
```diff
@@ -562,39 +716,35 @@ sjrk.storyTelling.server.testServerWithStorageDefs.getSavedStory = function (sto
...
-        if (fileOptions) {
-            var exists = fs.existsSync(fileOptions.expectedUploadDirectory
-                + binaryRenameMap[testStoryModel.content[0][fileOptions.urlProp]]);
-
-            jqUnit.assertTrue("Uploaded file exists", exists);
-        }
-
         if (completionEvent && fileOptions) {
             completionEvent.fire(parsedData.content[0][fileOptions.urlProp]);
         }
```
cindyli, line 720: Shall the else block at line 751 to verify an error response when `authoringEnabled === false`?

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -131,16 +138,21 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             }
         ],
         "tags": ["cute", "cats"],
+        "timestampCreated": "",
```
cindyli, line 151: Don't forget to update `timestampCreated`, `timestampLastModified` and `timestampPublished` to proper values at proper stages.

src/ui/base-page-storyEdit.js
```diff
@@ -109,30 +115,17 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-            "onStoryShareRequested.submitStory": {
-                funcName: "sjrk.storyTelling.base.page.storyEdit.submitStory",
-                args: [
-                    "{storyEditor}.dom.storyEditorForm",
-                    "{storyPreviewer}.story.model",
-                    "{that}.events.onStoryShareComplete",
-                    "{that}.options.pageSetup.storyAutosaveKey"
-                ]
-            },
+            "onStorySaveToServerRequested.saveStoryToServer": "{that}.saveStoryToServer",
```
cindyli: `onStorySaveToServerRequested` to save the auto save to the server hasn't been called in this pull request. I guess it'll be in the later commits.

src/server/requestHandlers.js
BlueSlug: I've actually removed this part and refactored the function entirely, since the server was totally ignoring the try/catch and crashing anyway

src/server/middleware/saveStoryFile.js
BlueSlug: Indeed, I was thinking of passing in model detail containing the current filename in order to remove that from the server before operating on the new file. Since this is multi-part form data, it should be feasible.

I'd also considered creating the ID client-side, but decided to keep all ID generation on the server.

src/ui/uploader.js
BlueSlug: It was formerly used quite heavily in the `binaryRenameMap` collections that were passed around in order to associate the various story blocks with the files that were uploaded en masse on story publication. It's entirely possible that it isn't needed any more.

It seems it is only used in `testServerWithStorage.js`, now. I'll try to remove it and see if anything breaks :)

tests/server/testServerWithStorage.js
BlueSlug: I agree that it's a little bit spaghetti, with each test feeding into the next. Breaking it into sequences would also allow for more variety in the test cases.

Doing this will take a fair amount of time, I believe, so we may want to revisit the release date in that case.

tests/ui/js/base-page-storyEditTests.js
BlueSlug: Good catch, I'd forgotten about these three :) Thanks!

src/ui/base-page-storyEdit.js
BlueSlug: I added these three events together, but my plan for each of them has changed a little. I don't anticipate using `onStorySaveToServerRequested` in my current work for this Jira. Should I remove it, in that case?

cindyli: I think it's a good idea to remove it and only add when it'll be used. Thanks.

src/server/requestHandlers.js
```diff
@@ -199,20 +172,18 @@ sjrk.storyTelling.server.handleSaveStoryWithBinaries = function (request, dataSo
...
  * @param {Component} dataSource - an instance of sjrk.storyTelling.server.dataSource.couch.story
- * @param {Object.<String, String>} binaryRenameMap - a map of uploaded file names to paths
  * @param {Object} storyModel - a single story's model
  * @param {Object} successEvent - an infusion event to fire upon successful completion
  * @param {Object} failureEvent - an infusion event to fire on failure
  */
-sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenameMap, storyModel, successEvent, failureEvent) {
-    var id = uuidv1();
+sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, storyModel, successEvent, failureEvent) {
+    var id = storyModel && storyModel.id ? storyModel.id : uuidv4();
```
jobara: You could also write this as `fluid.get(storyModel, "id") || uuidv4()`

src/server/requestHandlers.js
```diff
@@ -222,45 +193,85 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
+                message: "Error retrieving story with ID " + id,
+                error: err
+            });
+        });
+    } else {
+        request.events.onError.fire({
+            isError: true,
+            message: "Saving is currently disabled."
+        });
+    }
```
jobara: It might be easier to read this if you move it to the top and just have a simple if condition that returns. 

```javascript
if (!authoringEnabled) { 
    request.events.onError.fire({
        isError: true,
        message: "Saving is currently disabled."
    });
    return;  
}
```

src/server/requestHandlers.js
```diff
@@ -222,45 +193,85 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
-                // Look for the uploaded file matching this block
-                var mediaFile = fluid.find_if(files, function (file) {
-                    return file.originalname === block.fileDetails.name;
-                });
-
-                // If we find a match, update the media URL. If not, clear it.
-                if (mediaFile) {
-                    sjrk.storyTelling.server.setMediaBlockUrl(block, mediaFile.filename);
-                    binaryRenameMap[mediaFile.originalname] = mediaFile.filename;
+sjrk.storyTelling.server.handleSaveStoryFile = function (request, dataSource, authoringEnabled) {
```
jobara: I'm not sure I'm following the workflow of this function. It says it handles saving, but I'm not exactly sure where that happens. It seems that it is already saved by the time this handler executes and that this function will attempt to rotate the image as needed and also verifies the save. Is that correct?

src/server/middleware/removePreviousStoryFile.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {Object} request - the incoming request
+ * @param {String} uploadedFilesDirectory - the directory where uploaded files are stored
+ */
+sjrk.storyTelling.server.middleware.removePreviousStoryFile.handle = function (request, uploadedFilesDirectory) {
+    if (request.req.body.previousFileUrl) {
+        var fileToDelete = path.join(uploadedFilesDirectory, path.basename(request.req.body.previousFileUrl));
+
+        fluid.log(fluid.logLevel.WARN, "Deleting file: " + fileToDelete);
+
+        if (fs.existsSync(fileToDelete)) {
```
cindyli: In the case the previous file doesn't exist, might be helpful to log this message.

When the file does get removed successfully, log the success too.

src/server/middleware/removePreviousStoryFile.js
```diff
@@ -0,0 +1,49 @@
...
+
+var fluid = require("infusion");
+var path = require("path");
+var fs = require("fs");
+require("kettle");
+
+var sjrk = fluid.registerNamespace("sjrk");
+
+// Middleware to remove the previous file associated with a story block
+fluid.defaults("sjrk.storyTelling.server.middleware.removePreviousStoryFile", {
```
jobara: Consider renaming, as it appears that this can remove any file, not just the previous one.

src/server/requestHandlers.js
```diff
@@ -111,50 +111,43 @@ fluid.defaults("sjrk.storyTelling.server.getStoryHandler", {
...
 
-        var responseAsJSON = JSON.stringify(response);
-        request.events.onSuccess.fire(responseAsJSON);
+            request.events.onError.fire({
+                isError: true,
+                message: noAccessErrorMessage
+            });
+        }
     }, function (error) {
         var errorAsJSON = JSON.stringify(error);
```
cindyli: errorAsJSON -> errorInStr?

jobara: This seems to be misnamed as it's a string value, not JSON.

src/server/requestHandlers.js
```diff
@@ -168,29 +161,9 @@ fluid.defaults("sjrk.storyTelling.server.saveStoryWithBinariesHandler", {
  * @param {Component} dataSource - an instance of sjrk.storyTelling.server.dataSource.couch.story
  * @param {Boolean} authoringEnabled - a server-level flag to indicate whether authoring is enabled
  */
-sjrk.storyTelling.server.handleSaveStoryWithBinaries = function (request, dataSource, authoringEnabled) {
+sjrk.storyTelling.server.handleSaveStory = function (request, dataSource, authoringEnabled) {
```
jobara, line 159: Rather than having flags at the handler level for `authoringEnabled` we should modify the configuration, probably using context awareness, to not include configuration related to editing, when authoring is disabled.

src/ui/base-page-storyEdit.js
```diff
@@ -109,58 +114,98 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                args: [
+                    "{that}.options.pageSetup.storySaveUrl",
+                    "{storyEditor}.story",
+                    "{that}.options.pageSetup.storyIdPath",
+                    "{that}.options.pageSetup.storyAutoloadSourceName",
+                    "{that}.events.onStorySaveToServerError"
+                ]
+            },
+            // Saves the current story to the server
+            saveStoryToServer: {
```
cindyli: With the use of the `saveNewStoryToServer()` name, `saveStoryToServer()` gives an impression of saving an existing story to the server. But this function is to make an ajax call to save any story. It probably make sense to clarify the name such as `postStoryToServer`?

src/ui/base-page-storyEdit.js
```diff
@@ -384,55 +452,72 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+     * @param {Component} storyEditPage - an instance of `sjrk.storyTelling.base.page.storyEdit`
+     * @param {String} viewPageUrl - the URL for the story View page
+     * @param {Component} story - an instance of `sjrk.storyTelling.story`
+     * @param {publishSuccessCallback} successCallback - to call on successful publish
+     * @param {Object} errorEvent - the event to be fired in case of an error
+     */
+    sjrk.storyTelling.base.page.storyEdit.publishStory = function (storyEditPage, viewPageUrl, story, successCallback, errorEvent) {
+        // set the publish timestamp to now and the flag to "true"
+        story.applier.change("timestampPublished", new Date().toISOString());
+        story.applier.change("published", true);
```
cindyli: Consider to optimize the change request to:

```suggestion
story.applier.change("", {
    "timestampPublished": new Date().toISOString(),
    "published": true
});
```

The benefit is when there's a model listener listening to the root of the model, this listener won't be fired twice.

src/ui/blockUi-editor-audioBlockEditor.js
```diff
@@ -16,14 +16,24 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     // an editing interface for individual audio-type blocks
     fluid.defaults("sjrk.storyTelling.blockUi.editor.audioBlockEditor", {
         gradeNames: ["sjrk.storyTelling.blockUi.editor.mediaBlockEditor"],
+        selectors: {
+            previewPlaceholder: ".sjrkc-st-block-audio-preview-placeholder"
+        },
+        modelListeners: {
```
cindyli: Could you please add a code comment to explain why this model listener is required? Thanks.

src/ui/blockUi-editor.js
```diff
@@ -114,6 +118,145 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                method: "toggle",
+                args: ["{change}.value"],
+                namespace: "responseAreaVisibleChange"
+            },
+            uploadButtonDisabled: {
+                this: "{that}.dom.uploadButton",
+                method: "prop",
+                args: ["disabled", "{change}.value"],
+                namespace: "uploadButtonDisabledChange"
+            }
```
jobara, line 172: May be able to collapse this down into a single listener handler that listens to model changes on all of the specified paths and performs the necessary toggling based off the model change.

see: https://docs.fluidproject.org/infusion/development/ChangeApplierAPI.html#the-path-entry-in-a-model-listener-declaration

src/ui/blockUi-editor.js
```diff
@@ -114,6 +118,145 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                                        outputValue: {
+                                            previewVisible: false,
+                                            progressAreaVisible: true,
+                                            responseAreaVisible: false,
+                                            uploadButtonDisabled: true
+                                        }
+                                    },
+                                    "errorReceived": {
+                                        outputValue: {
+                                            previewVisible: true,
```
cindyli, line 233: Shall the preview be visible when the upload receives an error?

src/ui/blockUi-editor.js
```diff
@@ -114,6 +118,145 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                    model: {
+                        fileObjectUrl: "{block}.model.mediaUrl",
+                        storyId: "{editor}.storyId"
+                    },
+                    listeners: {
+                        "{editor}.events.onMediaUploadRequested": {
+                            func: "{that}.events.onFileSelectionRequested.fire",
+                            namespace: "fireSelectionForFileUpload"
+                        },
+                        "onUploadComplete": {
```
cindyli: onUploadComplete -> onUploadSuccess? To be consistent with "onUploadError".

src/ui/blockUi-editor.js
```diff
@@ -114,6 +118,145 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                    selectors: {
+                        fileInput: "{that}.container"
+                    },
+                    model: {
+                        fileObjectUrl: "{block}.model.mediaUrl",
+                        storyId: "{editor}.storyId"
+                    },
+                    listeners: {
+                        "{editor}.events.onMediaUploadRequested": {
+                            func: "{that}.events.onFileSelectionRequested.fire",
```
jobara: I don't think you need to add the `fire` segment at the end in the listener block.

src/ui/story.js
```diff
@@ -17,6 +17,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     fluid.defaults("sjrk.storyTelling.story", {
         gradeNames: ["fluid.modelComponent"],
         model: {
+            id: null, // to be filled in on initialization
```
cindyli: It would be useful to document the db structure and fields including descriptions.

src/ui/blockUi-timeBased.js
```diff
@@ -28,51 +28,51 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                "this": "{that}.dom.mediaPreview.0",
                 method: "addEventListener",
                 args: ["ended", "{that}.events.onMediaEnded.fire"]
             },
             "onTemplateReady.bindOnMediaPlay": {
-                "this": "{that}.dom.mediaPlayer.0",
+                "this": "{that}.dom.mediaPreview.0",
                 method: "addEventListener",
                 args: ["play", "{that}.events.onMediaPlay.fire"]
             },
```
jobara: This should probably be collapsed into a single function to relay the necessary events.

src/ui/story.js
```diff
@@ -7,7 +7,7 @@ You may obtain a copy of the BSD License at
 https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENSE.txt
 */
 
-/* global fluid */
+/* global fluid, sjrk */
```
jobara: You can add "sjrk" to the eslint config so that you don't need to specify it as a global in all the files.

src/ui/story.js
```diff
@@ -32,8 +33,37 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            tags: [],
+            published: false,
+            // These timestamps are intended to be in ISO-8601 format and UTC time
+            // https://en.wikipedia.org/wiki/ISO_8601
+            // E.g. "2020-04-22T16:50:44.324Z"
+            timestampCreated: "",
+            timestampLastModified: "",
+            timestampPublished: ""
+        },
+        invokers: {
```
cindyli: According to the discussion with @BlueSlug, `timestampLastModified` is created for addressing [SJRK-345 issue](https://issues.fluidproject.org/browse/SJRK-345) when we need to set up a rule to check staled stories. Since we haven't set up any rules in relation to this timestamp, it might be good to remove this code before it slips out of our mind. @BlueSlug will keep the commit hash to this code in SJRK-345 for the reference in the future when this code is useful.

src/ui/uploader.js
```diff
@@ -28,13 +28,20 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
-        // Holds a fileObjectURL for file preview purposes
         model: {
-            fileObjectURL: null,
-            fileDetails: null
+            // uploadState can be one of the following values:
+            // "ready" (the initial state), "uploading", "errorReceived"
+            uploadState: "ready",
+            fileObjectUrl: null, // for the file preview
+            previousFileObjectUrl: null, // for file deletion on change/upload
+            serverUploadUrl: "/stories/",
```
jobara: I don't think this should be a model value, as it is likely not something that would change, and isn't part of the "state".

src/server/serverSetup.js
```diff
@@ -164,11 +169,16 @@ fluid.defaults("sjrk.storyTelling.server.app.storyTellingHandlers", {
...
+        saveStoryHandler: {
+            type: "sjrk.storyTelling.server.saveStoryHandler",
             "route": "/stories/",
             "method": "post"
         },
+        saveStoryFileHandler: {
+            type: "sjrk.storyTelling.server.saveStoryFileHandler",
+            "route": "/stories/:id",
+            "method": "post"
+        },
```
jobara, line 188: Should review the end points for proper RESTful semantics. In this case POST to create a story makes sense. However uploading a file would likely be better to POST to /stories/:id/files. Updating specific files, or stories can use PUT and deleting a file or story can use DELETE.

src/ui/uploader.js
cindyli: Some thoughts from the discussion with @BlueSlug and @jobara:

1. The upload url and method should be configurable;
2. These values make more sense to be in the "members" option rather than "model".

src/server/serverSetup.js
jobara: Would also be good to have all of the story related files, that aren't in the database, stored in a directory named off the story id. This would make organization easier as well as deleting the story.

src/ui/uploader.js
```diff
@@ -28,13 +28,20 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
         },
-        // Holds a fileObjectURL for file preview purposes
         model: {
-            fileObjectURL: null,
-            fileDetails: null
+            // uploadState can be one of the following values:
+            // "ready" (the initial state), "uploading", "errorReceived"
+            uploadState: "ready",
+            fileObjectUrl: null, // for the file preview
+            previousFileObjectUrl: null, // for file deletion on change/upload
```
jobara: We can probably remove the need for the `previosFileObjectUrl` model path. Instead we can add a model listener for changes to the fileObjectURL and trigger a delete when it changes. Or there is a request to re-upload actually fire a delete immediately for the current fileObjectUrl if one exists. We would need to update the middleware that handles the upload. This will also make things more RESTful.

src/server/requestHandlers.js
BlueSlug: As we discussed, this is out of scope for this Jira. I have filed [SJRK-415](https://issues.fluidproject.org/browse/SJRK-415) to work on this improvement at a later date.

src/server/serverSetup.js
BlueSlug: As we discussed, the work to make the server API more closely follow RESTful semantics is out of scope for this Jira, so I've filed [SJRK-416](https://issues.fluidproject.org/browse/SJRK-416) to tackle this work.

The work to save files in directories named after the storyID rather than in the same directory with a filename based on the storyID still makes sense within the scope of SJRK-289.

src/ui/blockUi-editor.js
BlueSlug: The designs for [SJRK-396](https://issues.fluidproject.org/browse/SJRK-396) seem to suggest that the preview will be hidden when the upload receives an error. I will set the value to `false` for now, and we can review the question when SJRK-396 is picked up.

src/ui/story.js
BlueSlug: I began to work on this, but then realized it affects nearly every file (as `fluid` is [also specified](https://github.com/fluid-project/eslint-config-fluid/blob/041cd0f476fc02d0293914962c14912750e753f5/.eslintrc.json#L7) in the `.eslintrc.json` for `eslint-config-fluid` itself). Due to the already cluttered nature of this PR, I decided this is out of scope and described the work under the freshly-filed [SJRK-417](https://issues.fluidproject.org/browse/SJRK-417)

src/ui/blockUi-editor.js
BlueSlug: As we discussed earlier today, this would require a model listener that listens to multiple paths (which is possible) but which also needs to know which DOM element it is going to be acting on.

Short of making the names of the model values consistently refer to the selector names and automatically extracting the latter from the former (which strikes me as a brittle and overcomplicated solution), I'm not sure there's a nice way to combine these into a single listener.

src/ui/story.js
BlueSlug: I have added a comment with the commit hash on the SJRK-345 Jira

src/ui/blockUi-editor.js
jobara: Let's not worry about this for the current PR and we can revisit if needed later.

src/ui/uploader.js
BlueSlug: Would it be reasonable to count this as part of the work for [SJRK-416](https://issues.fluidproject.org/browse/SJRK-416)? If so, I can adjust the description of that Jira to explicitly include the file upload endpoints.

jobara: @BlueSlug is that because of the DELETE operation?

BlueSlug: That's right: since making this change will require changing the server endpoints, it might make sense to do it at that same time

src/ui/blockUi-editor-audioBlockEditor.js
BlueSlug: I tried to determine why this was only in place for the audio block editor, and pored over my notes and change logs from when I added this model listener (back on May 20), and it seems that at the time the audio block was the only type that required its own explicit model listener to hide and show the preview placeholder image (the little microphone icon) along with the preview element itself (in this case, the `audio` element).

Since then, and confirmed by the testing I've just done, video block preview placeholders aren't hiding with their `video` elements. Tangentially related to this, Chrome has started to exhibit the symptoms of [SJRK-226](https://issues.fluidproject.org/browse/SJRK-226), which previously only applied to Firefox. With this in mind, I have added this model listener for **all 3** media block types (`audio`, `image`, and `video`), and have also solved SJRK-226 by making the image block preview placeholder its own element (the same one shared by the other two types).

Furthermore, because the `imageBlockEditor` grade is currently not derived from the same base as the `audioBlockEditor` and `videoBlockEditor` grades (which are both instances of `mediaBlockEditor`), this solution required a bit of duplication. The work to combine these three block type editors into `mediaBlockEditor` is described in [SJRK-175](https://issues.fluidproject.org/browse/SJRK-175), which will hopefully get some attention in the future, but may be out of scope for this issue.

Please see the commit message for this commit for more details: https://github.com/fluid-project/sjrk-story-telling/pull/77/commits/c539dd0f0b3b9554b6e72affdd770d87ab3301b9

BlueSlug:
This PR is now ready for more review, and I've taken it out of "draft" mode. There are still some tests to add to address the deficit in test coverage that all the new code has introduced, but I doubt we'll have time to handle all of it, according to our schedule.

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "sjrk-story-telling",
-    "version": "0.3.0",
+    "version": "1.0.0",
```
jobara: I believe we've decided this will be an 0.4.0 release instead

src/server/middleware/deleteFile.js
```diff
@@ -0,0 +1,58 @@
...
+        fluid.log(fluid.logLevel.WARN, "Deleting file: ", fileToDelete, "...");
+
+        try {
+            if (fs.existsSync(fileToDelete)) {
+                fs.unlinkSync(fileToDelete);
+            } else {
+                throw "The specified file did not exist, and could not be deleted: " + fileToDelete;
+            }
+
+            fluid.log(fluid.logLevel.WARN, "Successfully deleted file: ", fileToDelete);
```
jobara: Seems like a log level of Info or Important would be sufficient here. 
https://docs.fluidproject.org/infusion/development/coreapi#fluidloglevel

src/server/middleware/deleteFile.js
```diff
@@ -0,0 +1,58 @@
...
+ * {@link https://github.com/fluid-project/kettle/blob/master/docs/Middleware.md}
+ *
+ * @param {Object} request - the incoming request
+ * @param {String} uploadedFilesDirectory - the directory where uploaded files are stored
+ */
+sjrk.storyTelling.server.middleware.deleteFile.handle = function (request, uploadedFilesDirectory) {
+    if (request.req.body.previousFileUrl) {
+        var fileToDelete = path.join(uploadedFilesDirectory, path.basename(request.req.body.previousFileUrl));
+
+        fluid.log(fluid.logLevel.WARN, "Deleting file: ", fileToDelete, "...");
```
jobara: Seems like a log level of Info or Important would be sufficient here. 
https://docs.fluidproject.org/infusion/development/coreapi#fluidloglevel

src/server/requestHandlers.js
```diff
@@ -111,50 +109,40 @@ fluid.defaults("sjrk.storyTelling.server.getStoryHandler", {
...
  * @param {Component} dataSource - an instance of sjrk.storyTelling.server.dataSource.couch.story
- * @param {String} uploadedFilesHandlerPath - the path on the server to uploaded files
  */
-sjrk.storyTelling.server.handleGetStory = function (request, dataSource, uploadedFilesHandlerPath) {
+sjrk.storyTelling.server.handleGetStory = function (request, dataSource) {
     var id = request.req.params.id;
     var promise = dataSource.get({directStoryId: id});
 
-    promise.then(function (response) {
+    var noAccessErrorMessage = "An error occurred while retrieving the requested story";
```
jobara, line 117: We may need to review error messages and make them more configurable on the server. Perhaps a JIRA for the future.

src/server/requestHandlers.js
```diff
@@ -111,50 +109,40 @@ fluid.defaults("sjrk.storyTelling.server.getStoryHandler", {
...
-                    block.mediaUrl = uploadedFilesHandlerPath + "/" + block.mediaUrl;
-                }
-                return block;
-            }
-        });
+    promise.then(function (response) {
+        if (response.published) {
+            request.events.onSuccess.fire(JSON.stringify(response));
+        } else {
+            fluid.log("Unauthorized: cannot access an unpublished story: " + id);
```
jobara: Maybe this one should have a Warn level log.

src/server/requestHandlers.js
```diff
@@ -111,50 +109,40 @@ fluid.defaults("sjrk.storyTelling.server.getStoryHandler", {
...
-        var responseAsJSON = JSON.stringify(response);
-        request.events.onSuccess.fire(responseAsJSON);
+            request.events.onError.fire({
+                isError: true,
+                message: noAccessErrorMessage
+            });
+        }
     }, function (error) {
-        var errorAsJSON = JSON.stringify(error);
+        fluid.log("Error getting story with ID " + id + ", error detail: " + JSON.stringify(error));
```
jobara: This seems also like it should have a log level of Warn

src/server/requestHandlers.js
```diff
@@ -222,42 +188,85 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
-                    sjrk.storyTelling.server.setMediaBlockUrl(block, mediaFile.filename);
-                    binaryRenameMap[mediaFile.originalname] = mediaFile.filename;
-                } else {
-                    sjrk.storyTelling.server.setMediaBlockUrl(block, null);
-                }
+    var id = request.req.params.id;
+
+    // verify that the story exists and isn't published before continuing
+    dataSource.get({directStoryId: id}).then(function (story) {
+        if (story && !story.published && request.req.file) {
```
jobara: Is it possible for the `dataSource.get` to resolve without a story?

src/server/requestHandlers.js
```diff
@@ -222,42 +188,85 @@ sjrk.storyTelling.server.saveStoryToDatabase = function (dataSource, binaryRenam
...
+                request.events.onSuccess.fire(request.req.file.destination + "/" + request.req.file.filename);
             }
+        } else {
+            request.events.onError.fire({
+                isError: true,
+                message: "Error saving file: file was not provided or story is already published"
+            });
         }
+    }, function (err) {
+        request.events.onError.fire({
```
jobara, line 264: Are we returning proper [http response codes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) with all of the errors?

src/ui/base-page-storyEdit.js
```diff
@@ -274,50 +364,73 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+     * @param {Component} story - an instance of `sjrk.storyTelling.base.page.storyEdit`
+     * @param {Component} blockManager - the storyEditor's blockManager component
+     * @param {String} sourceName - the name of the Infusion change source for this update
+     */
+    sjrk.storyTelling.base.page.storyEdit.loadStoryContent = function (savedStoryData, story, blockManager, sourceName) {
+        story.applier.change("", savedStoryData, null, sourceName);
+
+        // build the blockUIs from the story content array
+        blockManager.createBlocksFromData(savedStoryData.content);
+    };
```
jobara: It seems this whole function can be removed. Instead you could make the related invoker declaratively trigger the changeApplier. You could then have a modelListener, which could only listen to a specific source if need be, to call `blockManager.createBlocksFromData`.

src/ui/base-page-storyEdit.js
```diff
@@ -334,6 +447,43 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+            counters.uploads--;
+
+            if (newState === "errorReceived") {
+                counters.errors++;
+            }
+        } else if (previousState === "ready") {
+            counters.uploads++;
+        }
+
+        storyEdit.applier.change(["uploadCounters"], counters);
```
jobara: Unless this is handled in a model relay, it doesn't appear that there are any limits on the range. At least we shouldn't allow the count to go below 0.

src/ui/base-page-storyEdit.js
```diff
@@ -384,55 +534,74 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+     * @param {String} viewPageUrl - the URL for the story View page
+     */
+
+    /**
+     * Sets a story to "published" and redirects the user to the new story page
+     *
+     * @param {Component} storyEditPage - an instance of `sjrk.storyTelling.base.page.storyEdit`
+     * @param {String} viewPageUrl - the URL for the story View page
+     * @param {Component} story - an instance of `sjrk.storyTelling.story`
+     * @param {publishSuccessCallback} successCallback - to call on successful publish
```
jobara: I wonder if this should just return a promise instead?

src/ui/blockUi-editor.js
```diff
@@ -16,12 +16,34 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
             // reorder buttons are disabled depending on the block's order:
-            // if the block is first in the order, the "up" button is disabled
             // if the block is last in the order, the "down" button is disabled
-            moveBlockDownDisabled: "{block}.model.lastInOrder",
-            moveBlockUpDisabled: "{block}.model.firstInOrder"
+            blockLastInOrderToDownButtonDisabled: {
+                source: "{block}.model.lastInOrder",
+                target: "moveBlockDownDisabled",
+                singleTransform: {
+                    type: "fluid.transforms.value"
```
jobara: should this be `fluid.transforms.identity`?

src/ui/blockUi-timeBased.js
```diff
@@ -28,51 +28,42 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+                "args": ["{that}.dom.mediaPreview"]
             }
         },
+        // describes the relationship of DOM events to associated Infusion events
+        mediaPlayerEventMapping : {
+            canplay: "{that}.events.onMediaReady.fire",
+            durationchange: "{that}.events.onMediaDurationChange.fire",
+            ended: "{that}.events.onMediaEnded.fire",
+            play: "{that}.events.onMediaPlay.fire"
+        },
```
jobara: Another option here would be to just have the Infusion event names on the right hand side. In that case `sjrk.storyTelling.blockUi.timeBased.registerMediaPlayerEvents` would need to take in `that` and would construct the output even firer like `that.events[eventName].fire'

src/ui/uploader.js
```diff
@@ -44,98 +54,157 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+                processData : false,
+                type        : uploadOptions.uploadMethod,
+                timeout     : uploadOptions.uploadTimeout,
+                success     : function (data, textStatus, jqXHR) {
+                    fluid.log(jqXHR, textStatus);
+
+                    uploadOptions.completionEvent.fire(data);
+                },
+                error       : function (jqXHR, textStatus, errorThrown) {
+                    fluid.log(jqXHR, textStatus, errorThrown);
```
jobara: maybe this one should be have a Warn level

src/ui/uploader.js
```diff
@@ -44,98 +54,157 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
-            URL.revokeObjectURL(that.model.fileObjectURL);
-            that.applier.change("fileObjectURL", URL.createObjectURL(currentFile));
-            that.applier.change("fileDetails", fileDetails);
-        } else {
-            that.applier.change("fileObjectURL", "");
-            that.applier.change("fileDetails", "");
-        }
+    sjrk.storyTelling.block.singleFileUploader.updateFileObjectInformation = function (that) {
+        that.applier.change("previousFileObjectUrl", that.model.fileObjectUrl);
+        that.applier.change("fileObjectUrl", "");
```
jobara: This could be done declaratively and in a single change request.

src/server/requestHandlers.js
BlueSlug: I agree, and they should be localizable. This goes for the logging and messages in the rest of the request handlers and middleware, as well. I have filed [SJRK-419](https://issues.fluidproject.org/browse/SJRK-419) to tackle this work.

src/server/middleware/deleteFile.js
BlueSlug: Info seems like the most reasonable choice, after all, as this would fall into the category of "corresponding to verbose debugging messages", especially considering this will be logged any time an author uploads a new file for a given block

src/server/requestHandlers.js
BlueSlug: Warn seems reasonable here, though if we get a large number of attempts to access stories without authorization, then we may want to drop this to "Important" if the logs are getting too big

BlueSlug: We should probably sort out log backups first ;P

src/server/requestHandlers.js
BlueSlug: I believe this won't ever be the case, so the check for `story` is probably overkill. Removing it for now. If we discover a case where it *is* possible, then we'll want to add the same check to `sjrk.storyTelling.server.handleGetStory`

src/server/requestHandlers.js
BlueSlug: We aren't, currently. I've been following [the example set by `fluid.promise`](https://docs.fluidproject.org/infusion/development/promisesapi#promiserejecterror) and returning `isError: true` and a `message` string with every error, but HTTP status codes would be good to add as well. Would you like this as part of this work, or should we file a Jira to tackle it project-wide?

src/ui/blockUi-editor.js
BlueSlug: These two functions are aliases of one another, and so I chose to use `value` since it's possibly easier to understand at a glance. That said, an Infusion developer would be used to seeing `identity`, and we do use the latter in our tests. Do you have a preference?

src/ui/base-page-storyEdit.js
BlueSlug: I agree with this change and partially implemented it, but I've run into trouble setting up the model listener.

I have tried:
1. setting up the model listener on the `storyEdit` grade (top-level, where the original `loadStoryContent` invoker is)
2. moving the model listener to the `storyEditor` component (mid-level, it owns the `blockManager` and `story` components)
3. moving the model listener to the `storyEditor`'s story component (the model being listened to)
4. moving the model listener to the `blockManager` (whose invoker is called by the listener) and setting its key (i.e. path) to `"{story}.model"`
5. moving the model listener to the `blockManager` and setting its key to `"{storyEditor}.story.model"`
6. moving the model listener to the `blockManager` and setting the `segs` and `context` keys to `[]` and `"story"`, respectively

Solutions 1, 2 and 3 don't seem to be working due to timing issues. The `blockManager` component has a `createOnEvent` defined for it, and so when the `story` is being constructed, `blockManager` may not exist yet. That said, it should exist when `storyEdit` is being constructed, so this is surprising. Here is the error message in one of those cases, showing the definition of the model listener:
```
"Assertion failure - check console for more details: Error in listener record - could not resolve reference
{
    "func": "{storyEditor}.blockManager.createBlocksFromData",
    "args": ["content"],
    "includeSource": ["storyAutoload"],
    "namespace": "createEditorBlocksFromStoryData"
}
to a listener or firer. Did you miss out "events." when referring to an event firer?"
```

Solutions 3, 4 and 5 don't work either, though no errors are thrown and it fails silently.

src/ui/base-page-storyEdit.js
BlueSlug: I don't think it's possible for the counters to go below 0, as neither of them is decremented when the `previousState` is not one of the three cases handled by the outer `if` block. Assuming model relays & the `changeApplier` are synchronous, i.e. they won't be able to run simultaneously, then the same counter could not be decremented without a corresponding initial increment. I can add the following just before the `changeApplier` call, if you think it would be helpful:
```JavaScript
// ensure the counters are not negative
counters.uploads = Math.max(counters.uploads, 0);
counters.errors = Math.max(counters.errors, 0);
```

There may also be a way to achieve the aims of this function (and enforce its minima) more elegantly via model relays and transform functions.

src/server/requestHandlers.js
jobara: Let's file this as a new JIRA and review things as a whole.

src/server/middleware/deleteFile.js
```diff
@@ -0,0 +1,58 @@
...
+            if (fs.existsSync(fileToDelete)) {
+                fs.unlinkSync(fileToDelete);
+            } else {
+                throw "The specified file did not exist, and could not be deleted: " + fileToDelete;
+            }
+
+            fluid.log(fluid.logLevel.INFO, "Successfully deleted file: ", fileToDelete);
+        } catch (err) {
+            fluid.log(fluid.logLevel.FAIL, "Error deleting file \"", fileToDelete, "\".");
+            fluid.log(fluid.logLevel.FAIL, "Error detail: ", err.toString());
```
jobara: Are these really Fail level? Will the system terminate shortly after this?

src/ui/base-page-storyEdit.js
```diff
@@ -274,50 +364,73 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
-            if (block.imageUrl) {
-                block.imageUrl = null;
-            } else if (block.mediaUrl) {
-                block.mediaUrl = null;
-            }
+    sjrk.storyTelling.base.page.storyEdit.createNewStoryOnServer = function (storySaveUrl, story, storyIdPath, sourceName, errorEvent) {
+        // The "story created" moment is now
+        story.applier.change("timestampCreated", new Date().toISOString());
+
+        var serverSavePromise = sjrk.storyTelling.base.page.storyEdit.updateStoryOnServer(storySaveUrl, story.model);
```
jobara, line 408: What is in the `story.model` at this point? Is it just the `timestampCreated`? If so, I wonder if it would make more sense to have the server create the time stamp. In terms of the timestamp currently what happens if the `serverSavePromise` is rejected?

src/ui/base-page-storyEdit.js
jobara: I'm having a bit of trouble understanding who should be responsible for which parts. It looks like this is a method on `sjrk.storyTelling.base.page.storyEdit` which modifies the model of the `storyEditor` subcomponent's `story` subcomponent's model. It then makes a call to the `storyEditor`'s `blockManager`'s `createBlocksFromData` method. I'm not entirely clear on the distinction between `storyEdit` and `storyEditor`. I'm guessing the former is the page, and the latter is just the portion where stories are written. Is that correct?

jobara: I think I have this working with the attached changes. However, I'm still not sure of the factoring. 
[SJRK-289.patch.zip](https://github.com/fluid-project/sjrk-story-telling/files/5517903/SJRK-289.patch.zip)

src/ui/base-page-storyEdit.js
```diff
@@ -334,6 +447,40 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+     * @param {String} previousState - the previous state of the value
+     * @param {String} newState - the new state of the value
+     * @param {Component} storyEdit - an instance of `sjrk.storyTelling.base.page.storyEdit`
+     */
+    sjrk.storyTelling.base.page.storyEdit.updateUploadCounters = function (previousState, newState, storyEdit) {
+        var counters = fluid.copy(storyEdit.model.uploadCounters);
+
+        if (previousState === "errorReceived") {
+            counters.errors--;
+            counters.uploads++;
```
jobara: Is it not possible to go from "errorReceived" to "ready"?

src/ui/blockUi-editor.js
jobara: identity might just be clearer about the expectations of what will happen, which is to say, no model transformation.

src/ui/base-page-storyEdit.js
BlueSlug: As we discussed at our daily check-in today, this is possible when the `resetUploadState` invoker is called, which is currently only on block deletion. I'll add a check for it in this `if` block

src/ui/base-page-storyEdit.js
BlueSlug: That's correct: `storyEdit` is the page which includes the `storyEditor`, `storyViewer` and a couple other `ui` grade subcomponents

`storyEditor` is the `ui` component which has the story (which in turn has the data), a templateManager which renders everything, the binder, and the blockManager (which manages the blockUi components which each in turn have a template manager & block)

This patch worked, thanks! I've opted to use the previous event source value (`storyAutoLoad`) which was already present, rather than the new one in your path (`loadStoryContent`, named after the invoker that initiates the change). The reason for this is that the former is the source for the changes triggered in `sjrk.storyTelling.base.page.storyEdit.createNewStoryOnServer`, and we don't need/want to call `createBlocksFromData` at this point.

src/ui/base-page-storyEdit.js
BlueSlug: I think, given that it's called from an invoker which is in turn called in an event listener, I'm not sure how a return value could be accessed.

To stay consistent with the rest of the project, it might make more sense to pass in an event to fire which can return a promise, and then a listener to that event can go from there.

BlueSlug: I've opted to add an `onStoryPublishSuccess` event that returns the story ID, which allowed me to simplify the function a bit and use listeners to do the rest. I think it's worth revisiting the idea of how the rest of the system could be adapted if `publishStory` returns a promise, though I'm not sure at this point what that would look like

src/ui/base-page-storyEdit.js
```diff
@@ -26,6 +26,17 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
                         namespace: "uploadStateToCounters"
                     }
                 }
+            },
+            "blockManager.loadStoryContentIntoBlockUIs": {
+                record: {
+                    "{storyEditor}.story.model": {
+                        func: "{that}.createBlocksFromData",
```
jobara: Realizing it's a bit ambiguous to use `{that}` here. I think we should update this to use `{blockManager}` or some other appropriate name that is more explicit.

jobara:
Due to the deadline for releasing these features we will need to merge without having finished writing tests, and thus will have reduced code coverage.

jobara:
merged at 125f245785f2b76487d168d6a39908dd05ed788a

PR #78, opened 2020-06-02
SJRK-361: Pin Node.js version in Dockerfile

gtirloni:
With GitHub Actions, I cannot force a build without creating a commit. In Jenkins, we'd just push the build button to re-run a job.

That means that whenever there's an update to Node.js, we don't have a way to update the actual containers that are running with the new image because we're using `FROM node:10-alpine`.

I'm proposing we pin the Node.js version here so every new version requires a PR/commit that will in turn trigger the CI/CD job that will deploy new containers.

I think this will have to be our strategy for Docker images going forward.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10-alpine
+FROM node:10.21.0-alpine
 
 WORKDIR /usr/src/app
 


cindyli:
@gtirloni, can you create a JIRA for this issue and link the pull request with the JIRA number. The format of commit messages in SJRK still needs to be prefixed with the JIRA number. Thanks.

jobara:
@gtirloni I'm not sure if this is what you're looking for/need, but if you go to GitHub Actions and open a workflow run, you are able to re-run it. See the `Re-run jobs` button in the top right corner ( https://github.com/fluid-project/sjrk-story-telling/runs/732371841?check_suite_focus=true )

gtirloni:
@jobara that's nice, I thought it was only for failed builds. 

I still think it'll be better to pin the Node.js version and have a record of changes in the code. Full GitOps :)

amb26:
+1 for pinned versions of everything :)

PR #84, opened 2020-06-10
SJRK-288 & SJRK-359: added block reordering

BlueSlug:
This work addresses:
- [SJRK-288](https://issues.fluidproject.org/browse/SJRK-288) (block reordering with buttons)
- [SJRK-359](https://issues.fluidproject.org/browse/SJRK-359) (block reordering with drag & drop)
- [SJRK-365](https://issues.fluidproject.org/browse/SJRK-365) (focus on whole block groups with forward and backward keyboard navigation)


src/ui/ui-storyEditor.js
```diff
@@ -153,6 +153,19 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
             blockManager: {
                 container: "{storyEditor}.options.selectors.storyEditorContent",
                 options: {
+                    distributeOptions: {
```
cindyli, line 145: I wonder what's the reason to use `distributeOptions` to distribute listeners for `onMoveBlockDown` and `onMoveBlockUp` events rather than directly defining them on `sjrk.storyTelling.blockUi.editor` component? Is it because you only want to attach these listeners to the story editor, not the story viewer? But is there a chance that up/down buttons are still available and clickable on the story viewer?

I feel it's less confusing to define listeners along with corresponding events if possible.

tests/ui/js/ui-storyEditorTests.js
```diff
@@ -146,30 +153,113 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                    element: "{storyEditor}.model.lastAddedBlock.dom.moveBlockDownButton"
+                },
+                {
+                    event: "{storyEditor}.reorderer.events.onMove",
+                    listener: "jqUnit.assert",
+                    args: ["Reorderer onMove fired on moveBlockDown button clicked"]
+                },
+                {
+                    // Call reorderBlockUp, listen for reorderer onMove
+                    func: "{storyEditor}.reorderer.reorderBlockUp",
```
cindyli: Is it necessary to call `reorderer.reorderBlockUp()` in the test? Shall it already be handled by the `onMoveBlockDown` listener when the down button is clicked at line 224?

I think this test sequence is not clear as it also manually triggers a couple of internal actions such as calling this `reorderBlockUp()` function and `reorderer.container.focusin` later. 

This test sequence can be simplified to: add blocks -> click "down" button -> the reorderer is refreshed -> the block is moved down to the proper position -> click "up" button -> the reorderer is refreshed -> up to the proper position.

In the sequence, check the inactive button states if needed.

docs/GRADES-UI.md
```diff
@@ -50,7 +51,9 @@
 * `sjrk.storyTelling.ui.storyEditor` is an editing interface for a `story` (which it has as a component). It has a
   handlebars template associated with it called `storyEditor.handlebars`, and a `binder` to connect the title, author
   and keywords fields to their respective model values in the `story`. It makes use of a `dynamicViewComponentManager`
-  called **blockManager** to add blocks of varying types on demand.
+  called **blockManager** to add blocks of varying types on demand. The `reorderer` component is for reordering individual
+  blocks via buttons on each (see `sjrk.storyTelling.block.editor` above), using CTRL+UP or CTRL+DOWN or by dragging and
```
cindyli: It would be helpful to have a text on the interface to explain how to move blocks using keyboard. On Mac, the key combinations working for me are CTRL+SHIFT+UP or CTRL+SHIFT+DOWN.

cindyli:
The inactive button states for the "up" button on the first block and the "down" button on the last block are not implemented.

src/ui/ui-storyEditor.js
```diff
@@ -224,6 +237,49 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
                         }
                     }
                 }
+            },
+            reorderer: {
+                type: "fluid.reorderList",
```
amb26, line 226: Lovely and hilarious to see this back after a decade :)
:cake:

src/ui/blockUi-editor.js
```diff
@@ -17,8 +17,35 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     fluid.defaults("sjrk.storyTelling.blockUi.editor", {
         gradeNames: ["sjrk.storyTelling.blockUi"],
         selectors: {
+            moveBlockDownButton: ".sjrkc-st-reorderer-move-down",
+            moveBlockUpButton: ".sjrkc-st-reorderer-move-up",
             selectedCheckbox: ".sjrkc-st-block-selection-checkbox"
         },
+        events: {
+            onReadyToBind: null,
+            onBlockControlsBound: null,
```
amb26: Given this event is apparently only of use in test cases I suggest that its definition and binding be moved into a grade defined in the test cases themselves.

src/ui/ui-storyEditor.js
```diff
@@ -224,6 +237,49 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                        },
+                        reorderBlockUp: {
+                            func: "{that}.reorderBlock",
+                            args: ["{arguments}.0", fluid.direction.UP]
+                        }
+                    },
+                    listeners: {
+                        // We should be refreshing the reorderer on each new block
+                        // being added, but we can't reliably know when a block
+                        // has finished being rendered. It would also be possible
```
amb26: Why not - do we have any asynchronous rendering operations? If so, how do they work? Seems likely we should be able to tag the refresh operation onto them rather than apply this potentially expensive operation on every focus.

src/ui/blockUi-editor.js
```diff
@@ -17,8 +17,35 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            onMoveBlockDown: null,
```
amb26: I think a separate event for each direction is a little too bulky - we can bind onto a single event and supply it with the appropriate fluid.direction.* argument. This would eliminate the two invokers also - and finally we can just implement the actual "reorderBlock" activity as a listener, eliminating the last invoker.

src/ui/ui-storyEditor.js
amb26: See above suggestion - I think this linkage should be implemented by either forwarding one event firing to another or declaring one event as an alias to another (latter probably not recommended)

cindyli:
When using mouse to click on the heading or the text input fields of a block, the focus keeps jumping to focus on the entire block, probably the reorderer drag-drop handle. To reproduce:

1. Go to the story editor and add a block in any type;
2. Use mouse to click focus on the heading input field;
3. The focus jumps to the entire block.

It has no problem to use keyboard to tab focus on input fields.

This issue is reproducible in latest firefox, chrome and safari.

src/ui/ui-storyEditor.js
BlueSlug: The reason `distributeOptions` is used here is twofold:
- the `blockUi.editor` components are dynamically created, and I wasn't able to craft an IoC reference in the `reorderer` component that would resolve properly
- I didn't want to add any references to the `reorderer` component in the `blockUi.editor` grade, as the `reorderer` is specific to `storyEditor`

I think @amb26's suggestion of forwarding the event through to the `storyEditor` may work best. I'll see what I can come up with that also allows me to combine the two events (as suggested in the review for `blockUi-editor.js`)

BlueSlug: To answer @cindyli's question about whether we anticipate reordering will be used in the `storyViewer`: we do not

tests/ui/js/ui-storyEditorTests.js
BlueSlug: The call to the `refresh` invoker was to force a refresh after blocks are added and to ensure the invoker was operating properly. I agree that, as an integration test, it's redundant. What I propose instead is to move the jQueryTrigger of `focusin` to right after the blocks are added, that way the sequence more accurately simulates the user's interaction with the editor.

Because the blocks are added dynamically, we can't get away without refreshing the reorderer before clicking on the "down" button.

src/ui/blockUi-editor.js
BlueSlug: Given that it isn't even necessary for the test cases, I have decided to remove it altogether. I think it was leftover from a previous factoring.

src/ui/ui-storyEditor.js
BlueSlug: @jobara and I were exploring this, we even attempted to incorporate a `mutationObserver` into the `storyEditor` before that became too unwieldy, but we believe the issue boils down to [`gpii.handlebars.renderer`'s use](https://github.com/GPII/gpii-handlebars/blob/master/src/js/client/renderer.js#L15-L23) of jQuery functions to affect markup changes. In this case, we're using [`$.html()`](https://api.jquery.com/html/).

Since jQuery doesn't afford any way to know that it's done its thing, we can't reliably know when a newly-added blockUi is ready to be fiddled with, and therefore when it's safe to refresh the reorderer (this becomes an issue once a `grabHandle` is specified for a `movable`, something to do with the reorderer's `labeller` component, as I recall).

I've messaged @the-t-in-rtf to discuss how we might solve this issue, but I hadn't provided him with much detail and our workaround made it a less pressing matter.

src/ui/blockUi-editor.js
BlueSlug: This is an excellent suggestion, which I've implemented for the most part.

I'm hesitant to remove the final `reorderBlock` invoker, since leaving it in means there's another spot that a programmer can reach in to change things without having to know the details of `sjrk.storyTelling.ui.storyEditor.reorderBlock`'s arguments. This really only saves having to pass one more argument (you'd need a reference to the `reorderer` in order to call the invoker anyway), so it won't be a massive difference, which is why I'm on the fence.

src/ui/ui-storyEditor.js
amb26: I can't quite follow your explanation - $.html() is synchronous so it has "done it's thing" immediately?

src/ui/blockUi-editor.js
```diff
@@ -16,9 +16,63 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            // if the block is first in the order, the "up" button is disabled
+            // if the block is last in the order, the "down" button is disabled
+            moveBlockDownDisabled: "{block}.model.lastInOrder",
+            moveBlockUpDisabled: "{block}.model.firstInOrder"
+        },
+        modelListeners: {
+            "moveBlockDownDisabled": {
+                func: "{that}.setButtonDisabled",
+                args: ["{that}.dom.moveBlockDownButton", "{change}.value"],
+                excludeSource: ["init"],
```
amb26, line 39: Should be able to avoid partnering this with the "onReady" duplicate listeners, especially given you are on the FLUID-6145-era Infusion - what is the result if you simply remove the excludeSource annotation here?

src/ui/ui.js
```diff
@@ -205,4 +207,74 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+    /**
+     * Updates the order of each block model according to the order of the blockUi
+     * elements in the story content editing area
+     *
+     * @param {Component} that - an instance of sjrk.storyTelling.ui.storyUi
+     * @param {Object} completionEvent - the event to be fired upon successful completion
+     */
+    sjrk.storyTelling.ui.updateBlockOrder = function (that, completionEvent) {
+        var blockUis = that.reorderer.container.children();
+        var loopCounter = 0;
```
amb26: This odd way of structuring the iteration will likely throw off a reader - why not write a conventional "for" loop iterating over blockUIs, or else write blockUIs.forEach(function (blockUi, blockIndex) { .... ?

src/ui/ui.js
```diff
@@ -205,4 +207,74 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+        var numberOfBlocks = blockUis.length;
+
+        // used to retrieve the class name for each block
+        // within the blockManager's managedViewComponentRegistry
+        var managedClassNamePattern = sjrk.storyTelling.ui.getManagedClassNamePattern(that.blockManager);
+
+        fluid.each(blockUis, function (blockUi) {
+            var managedClassName = fluid.find(blockUi.classList, function (value) {
```
amb26: I haven't traced through all the details of this, but is the predicate here attempting to have the effect of https://api.jquery.com/hasClass/ ?

src/ui/ui-storyEditor.js
BlueSlug: I believe I misunderstood the root cause of the issue. In the case that we call the `refresh` invoker after adding a new block, the problem would arise when we attempt to add many blocks very quickly, either on loading an autosave or simply clicking the add button very rapidly. The reorderer will attempt to refresh itself, and upon coming into contact with a block that has a container added but not actually rendered, we'll get the error.

@jobara and I have discussed implementing a kind of event queue to manage adding blocks one at a time and waiting until the queue is empty to call refresh

src/ui/blockUi-editor.js
BlueSlug: I believe the difference is invisible to the user. The model values will be set early on when the DOM isn't ready (the buttons aren't rendered yet), so nothing happens when `.prop()` is called. The listeners hanging off of `onReadyToBind` are to set the buttons once the markup is rendered, and the modelListeners will handle any subsequent changes. I'm ambivalent about excluding init, at this point, though I suppose we might have some barely-perceptible performance gains by not handling it on init.

Can you please elaborate on what FLUID-6145 is allowing me to do or not do?

src/ui/ui.js
BlueSlug: What I attempted to do is to get a unique class name from the collection that matched a certain pattern, and then feed that class name into the `blockManager`'s component registry to get a reference to the `blockUi.editor` component that matches that class name, in order to reach its `block` subcomponent. I effectively wrote startsWith, though I think that's not supported in IE. Whether we're still supporting IE is something I'm unsure of.

I don't think `hasClass()` does what I want (I could be wrong), but if you know of any other approaches that fit the bill I would be glad to learn!

jobara: Infusion is no longer explicitly supporting IE. We haven't removed code for it, but will no longer be testing in IE or actively writing code to support it. That means that any new release (pre-release or stable) cannot be relied on to support IE. So it may be time to revisit and/or define the browser support for Story Telling.

jobara: @BlueSlug if I'm following the logic of the function, what you want to do is to take the DOM position of an element, find what Infusion component corresponds to that element, and then update the model values of that component with the position index. So maybe the first question is to @amb26 about whether there is an easy way to find a component from its container.

Another question is about the approach as a whole. That is you seem to only need this to update the disabled/enabled state of the up/down arrows for reordering. Here are some thoughts to get your started, none of them may be good though: 
- You could continue with the DOM based approach and on reorder have the components inspect where they are in the parents list of children or maybe if they have a previous or next sibling. In general modelizing state is better. 
- Should a reorderable element care about it's position? Should a single block UI really have to know that it is reorderable and where it is in the list? Perhaps this reordering behaviour could be the responsibility of a parent component that sort of decorates its children with the ability to be reordered. And there by managing the state of the individual reorderable elements.

amb26: Code to maintain an index of viewComponent containers to their components can be adapted from the "viewMapper" in fluidViewDebugging.js here - https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/FluidViewDebugging.js#L576
You can simplify this code considerably since there is extra material here to account for the possibility that the view component lies in an iframe.
However, I agree with @jobara 's comment that responsibility for all list-based functionality should be centralised in a parent component.

colinbdclark: > Infusion is no longer explicitly supporting IE. We haven't removed code for it, but will no longer be testing in IE or actively writing code to support it. That means that any new release (pre-release or stable) cannot be relied on to support IE. So it may be time to revisit and/or define the browser support for Story Telling.

I see no reason to support IE in the Storytelling Tool at this point. We have so much to do, so anything we can do to simplify development seems like a good idea. Let's not spend time trying to support Internet Explorer.

src/ui/blockUi-editor.js
amb26: One of the main improvements in the FLUID-6148/FLUID-6145 revision ensures that the timing of the "init" model transaction is fully predictable with respect to a component's "onCreate" - that is, it always comes before it. This should mean that the function of your duplicate listeners in "onReadyToBind" on lines 65-74 should be handlable by the model listener here as part of "init" - especially if you can move whatever it is that the component needs to be "ready" into the component's resources - the only such thing should be a template loading, as far as I can see, since everything following that should be synchronous

BlueSlug: We discussed this issue a bit [in IRC](http://irc-logs.fluidproject.org/%23fluid-work/%23fluid-work.2020-06-18.log) today, and came to the decision to defer this work for another Jira. I've filed [SJRK-364](https://issues.fluidproject.org/browse/SJRK-364) to capture your suggestions, please feel free to amend it if it isn't accurate :)

I'll be adding a comment in this part of the code linking to both this review item and the new Jira.

BlueSlug:
@jobara's fix in 3038c83 solves SJRK-365

src/ui/ui-storyEditor.js
cindyli: Bump. Please update the code comment with the root cause, a TODO and the proper fix you suggested.

src/ui/ui.js
```diff
@@ -18,8 +18,13 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
     fluid.defaults("sjrk.storyTelling.ui", {
         gradeNames: ["fluid.viewComponent"],
         events: {
-            onReadyToBind: null,
-            onControlsBound: null
+            onTemplateRendered: null,
```
cindyli: I wonder what's the point to introduce the new `onTemplateRendered` event if it's an equivalent of `onReadyToBind`?

src/ui/ui.js
```diff
@@ -116,8 +61,17 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
             storyTags: ".sjrkc-st-story-tags"
         },
+        // The name of the model change source when block order is updated.
+        // This should match the source specified in the block modelListeners
+        orderUpdateSource: "orderUpdate",
         events: {
-            onStoryUpdatedFromBlocks: null
+            onStoryUpdatedFromBlocks: null,
+            onBlockManagerCreated: null,
+            onReadyToBind: {
```
cindyli: Following the comment above, is your intention of using `onReadyToBind` event is to use it as an aggregate event for `onBlockManagerCreated` + `onTemplateRendered`? But as `onTemplateRendered` is the CreateOnEvent for instantiating `blockManager`, `onBlockManagerCreated` event should always be fired after `onTemplateRendered`. It doesn't seem making sense to include `onTemplateRendered` in the aggregation.

src/ui/ui.js
```diff
@@ -162,6 +117,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
                         }
                     },
                     listeners: {
+                        "onCreate.escalate": "{ui}.events.onBlockManagerCreated",
```
cindyli: Shall `onCreate.escalate` be triggered after `onCreate.createBlocksFromData`? Depending on the answer to the question above, if `storyUi.events.onReadyToBind` doesn't need to be an aggregate event combined with `onTemplateRendered`, this escalation can directly goes to `{ui}.events.onReadyToBind` and remove the intervening `onBlockManagerCreated` event.

src/ui/ui.js
```diff
@@ -205,4 +161,64 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+     * model values from the event merged successfully with the base values
+     * of the block
+     *
+     * @param {Object} modelValuesFromEvent - the model values to use
+     *
+     * @return {String} - the new grade's name
+     */
+    sjrk.storyTelling.ui.storyUi.getBlockGradeFromEventModelValues = function (modelValuesFromEvent) {
+        var gradeName = "sjrk.storyTelling.block-" + fluid.allocateGuid();
+        fluid.defaults(gradeName, {
```
cindyli, line 165: I'm a bit surprised that this grade declaration doesn't require a `gradeNames` such as:
```
fluid.defaults(gradeName, {
    ...
```

src/ui/ui.js
```diff
@@ -205,4 +161,64 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            // TODO: this should test that modelValuesFromEvent is a legitimate
```
cindyli, line 166: This comment sentence is incomplete.

src/ui/ui-storyEditor.js
```diff
@@ -68,14 +68,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
             onRemoveBlocksCompleted: null,
-            onEditorTemplateRendered: null,
-            onBlockManagerCreated: null,
-            onReadyToBind: {
-                events: {
-                    onEditorTemplateRendered: "{that}.events.onEditorTemplateRendered",
-                    onBlockManagerCreated: "{that}.events.onBlockManagerCreated"
-                }
-            }
+            onBlockOrderUpdated: null
```
cindyli: It seems `updateBlockOrder` is only used in the test. I wonder instead of listening to this event in the test, would you be able to test if the story blocks have been updated/sorted correctly?

tests/ui/js/ui-storyEditorTests.js
```diff
@@ -146,30 +153,103 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+                    element: "{storyEditor}.model.lastAddedBlock.dom.moveBlockDownButton"
+                },
+                {
+                    event: "{storyEditor}.reorderer.events.onMove",
+                    listener: "jqUnit.assert",
+                    args: ["Reorderer onMove fired on moveBlockDown button clicked"]
+                },
+                {
+                    // Refresh the reorderer, listen for onBlockOrderUpdated after it's done
+                    func: "{storyEditor}.reorderer.refresh"
```
cindyli: Do you need to manually trigger the `reorderer.refresh()`? Shouldn't it be triggered automatically when items are reordered?

tests/ui/js/ui-storyEditorTests.js
```diff
@@ -225,6 +305,92 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            bLessThanA: {
+                content: [{ id: "a", order: 1 }, { id: "b", order: 0 }],
+                expectedContent: [{ id: "b", order: 0 }, { id: "a", order: 1 }]
+            },
+            aSameAsB: {
+                content: [{ id: "a", order: 0 }, { id: "b", order: 0 }],
+                expectedContent: [{ id: "a", order: 0 }, { id: "b", order: 0 }]
+            },
+            bothFalse: {
+                content: [{ id: "a", order: false }, { id: "b", order: false }],
```
cindyli, line 379: what if one is `true` and the other is `false`? Thanks for writing such a comprehensive coverage. It reminds me that the sorting probably should return without proceeding if the `order` value is not a valid number. If a non numerical value reaches this function, something goes very wrong. ;)

cindyli:
The focus styling for "move up/down" buttons are quite vague and hard to notice. The focus styling for other buttons: add blocks, delete and update, has the same issue. If I remember correct, missing hover styling for all buttons is a known issue that has a separate ticket.

cindyli:
The focus styling for "move up/down" buttons on UIO THEMES are totally unnoticeable.

cindyli:
When dragging the mouse to select some buttons, the "play" button from the text-to-speech shows up. Clicking on it doesn't read anything. If this is a problem, it probably should be addressed in a separate JIRA.

Screen Shot 2020-06-23 at 9 20 53 PM

cindyli:
Refreshing the story editor replicates existing blocks:

1. Go to the story editor and add one block;
2. Enter some text in the block;
3. Refresh the page using CTRL+r or browser refresh button;
4. Now there are 2 blocks in the editor. The 2nd one is a replica of the first one with the same text.

Expected: There should always be same blocks in the editor before and after the refresh.

src/ui/ui.js
BlueSlug: In this case, the `onReadyToBind` event is being set up as a compound event, to which other sub-events can be added in implementing grades. I needed to add the definition of `onTemplateRendered` at this level in order to refer to it within `onReadyToBind`.

src/ui/ui.js
BlueSlug: Yes, the intent is to use it as an aggregate event. I also noticed that `onBlockManagerCreated` would always fire after `onTemplateRendered` for `ui` grades with a `blockManager` component (instances of `storyUi`), but the intent is to keep `onReadyToBind` from firing the moment the template is done rendering (which was previously the case in `ui`), as that is too early for the `storyEditor` grade.

I thought this solution was a wiser choice than overriding the listener in the `templateManager` component of `ui`, and it allows for a bit more flexibility in configuring the timing. It might also be premature optimization :) Would you like me to revert to the previous setup and find another way to delay `onReadyToBind` for the `storyEditor`?

cindyli: It don't seem like a good idea to share the same `onReadyToBind` between `ui` and `storyUI` if one is dependent on the other and they are fired in sequence. I would suggest to use different events, such as `onUIReady` for `ui` component and `onStoryUIReady` for `storyUI` component.

BlueSlug: Okay, I'll separate the usage of them. I may still leave the template render event name as `onTemplateRendered`, since that's more accurate, and limit `onReadyToBind` usage to `storyUi` grades

cindyli: Feel free to use names that make sense. I've never been good at coming up with names.

src/ui/ui.js
BlueSlug: The rest of the comment is on the next line. Should I restyle it somehow or combine it into one line? This should probably be a new Jira, actually

cindyli: Leave as is as long as it makes sense...

src/ui/ui.js
BlueSlug: Me too! The result of this function is passed into a `gradeNames` key in the `block` subcomponents of the `blockManager`, for what it's worth. The exact details require some careful parsing of the `blockManager` component and the grade it's based on, `sjrk.dynamicViewComponentManager`.

It is a weird setup, but it's the one we figured out that lets us dynamically add viewComponents with varying grades.

BlueSlug:
> When dragging the mouse to select some buttons, the "play" button from the text-to-speech shows up. Clicking on it doesn't read anything. If this is a problem, it probably should be addressed in a separate JIRA.
> 
> Screen Shot 2020-06-23 at 9 20 53 PM

We still haven't patched in the fix for FLUID-6512, which would eliminate the page reload when `Play` is clicked, but I'm not sure it's a problem that an empty selection isn't readying anything. The Orator is not designed to read anything other than what is displayed on the screen, if I recally correctly, so no alt-text or aria things. @jobara may be able to confirm this behaviour.

src/ui/ui.js
amb26: This isn't recommended since in current Infusion these defaults definitions will leak - whilst dynamically generated grades are used from time to time, this is only where there is no alternative. It seems like a better approach would be to ensure that the values fired in the event are already properly defaulted (in createBlocksFromData ?) and then to write the dynamic references to the event arguments inline in the definition of the "block" component.
General notes - 
- yes, supplying no grades at all in "mixin" grades is a reasonable technique, and was even sometimes required in the old framework where it was possible that trying to mix in "fluid.modelComponent" to a view component again would trash its argument list. This bug got fixed in the new framework.
- the definition of "storyUi" is too large and should be broken up at least at one point, perhaps at "blockManager", "managedViewComponents", or both. It would be helpful to supply at least some kind of base grade for "managedViewComponents" in any case to guide the reader as to what kind of material might get created.

BlueSlug: I agree with these suggestions (and thank you for them!), though I feel they're out of scope for this Jira.

I went to file a Jira to capture your feedback, but there are in fact four(!) that already address reviewing the DVCM and its usage as the blockManager, and hopefully the time to tackle them will come soon :)

The most relevant issues are [SJRK-115](https://issues.fluidproject.org/browse/SJRK-115) and [SJRK-126](https://issues.fluidproject.org/browse/SJRK-126), while [SJRK-61](https://issues.fluidproject.org/browse/SJRK-61) and [SJRK-262](https://issues.fluidproject.org/browse/SJRK-262) are a little more tangentially relevant.

src/ui/ui-storyEditor.js
BlueSlug: I began to work on this and added some explicit test cases that check the story contents before and after each reorder, but in doing so I surfaced a previously-unknown bug: it seems that there is something happening with the changeApplier call in `sjrk.storyTelling.ui.storyEditor.sortStoryContent` that is merging the contents of that value rather than replacing it. The result is that blocks in the `story`'s content array have nearly every possible key present, i.e. text blocks have their usual fields in addition to `imageUrl` **and** `mediaUrl`, `fileDetails`, `alternativeText`, etc., and this ends up in the database on publish.

I have tried, among other things:
- calling the changeApplier with "DELETE" on the `content` key before setting the new value
- bundling the individual block component order value updates as a single change transaction
- using `fluid.set` to set the `content` model value explicitly
- specifying the story grade's `mergePolicy` as `nomerge` (maybe this should have been `replace`...)

The consequences are minimal, basically only that the test story will look strange since certain block types will have unexpected fields, though we should find a solution before it causes any real problems.

amb26: The fact that it is "accumulating" is a well-known design feature of the ChangeApplier since its earliest days - whether it is the best choice or not is secondary to the issue that doing anything about it at this point would break thousands of lines of code depending on it.
The correct approach is a combination of your first and second points - as seen in utilities such as this one in the prefs framework store - https://github.com/fluid-project/infusion/blob/master/src/framework/preferences/js/Store.js#L245 - this is such a frequently used pattern that it should be a core framework utility.
Point 3, using fluid.set directly into the model must never be operated since it will corrupt the ChangeApplier's records, and point 4 is inadvisable since mergePolicies are generally framework internals that should only be used as a last resort - their operation is fragile and not terribly well documented.
I agree we should get rid of this event.

src/ui/ui.js
jobara: @BlueSlug if these issues will not be addressed in this work, it would be useful to track the related issues in comments with the relevant code.

jobara:
> > When dragging the mouse to select some buttons, the "play" button from the text-to-speech shows up. Clicking on it doesn't read anything. If this is a problem, it probably should be addressed in a separate JIRA.
> > Screen Shot 2020-06-23 at 9 20 53 PM
> 
> We still haven't patched in the fix for FLUID-6512, which would eliminate the page reload when `Play` is clicked, but I'm not sure it's a problem that an empty selection isn't readying anything. The Orator is not designed to read anything other than what is displayed on the screen, if I recally correctly, so no alt-text or aria things. @jobara may be able to confirm this behaviour.

The Orator only reads the visible text. I suppose you should be able to configure that behaviour although it would likely require some new functions be implemented to know how to parse the text. 

The play button shows up with a selection, if you are finding that it shows up even when no text is selected, perhaps a JIRA against the Orator is warranted.

src/ui/ui-storyEditor.js
BlueSlug: The solution turned out to be to add a `DELETE` operation before updating the story contents in `sjrk.storyTelling.ui.storyUi.updateStoryFromBlocks`, not in `sjrk.storyTelling.ui.storyEditor.sortStoryContent`. I have wrapped up all changes in transactions either way, in order to economise model changes and avoid any unnecessary firing of the block-to-story model listener.

BlueSlug:
Ready for review

src/ui/ui-storyEditor.js
```diff
@@ -252,4 +293,107 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+            var managedClassName = fluid.find(blockUis[i].classList, function (value) {
+                if (typeof value === "string") {
+                    var patternMatches = value.match(managedClassNamePattern);
+                    return patternMatches === null ? undefined : patternMatches.input;
+                }
+            });
+
+            var block = that.blockManager.managedViewComponentRegistry[managedClassName].block;
+
+            var firstInOrder = i === 0 ? true : false;
```
cindyli: Line 332-341 can be simplified to:

```
block.applier.change({
    path: "",
    value: {
        order: i,
        firstInOrder: i === 0 ? true : false;
        lastInOrder: i === blockUis.length - 1 ? true : false;
    }
});
```

tests/ui/html/ui-storyEditor-manual-Tests.html
```diff
@@ -0,0 +1,202 @@
+<html>
```
cindyli, line 1: Could you add a documentation to describe available manual tests, their goals and how to access, etc? Thanks.

cindyli:
The insertion line when reordering blocks disappears when a UIO theme is selected:

1. Go to the story editor and insert a couple of blocks;
2. Open UIO and select a theme;
3. Drag and drop a block to reorder it;
4. The insertion line that shows up to indicate the drop location doesn't show up.

src/ui/ui-storyEditor.js
BlueSlug: This doesn't seem to work: Editor tests are failing and it breaks the reorderer buttons. I'll continue to investigate.

cindyli: This form works: 

```
block.applier.change("", {
  order: i,
  firstInOrder: i === 0 ? true : false,
  lastInOrder: i === blockUis.length - 1 ? true : false
});
```

src/ui/ui-storyEditor.js
```diff
@@ -252,4 +293,107 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/master/LICENS
...
+     * elements in the story content editing area
+     *
+     * @param {Component} that - an instance of sjrk.storyTelling.ui.storyUi
+     */
+    sjrk.storyTelling.ui.storyEditor.updateBlockOrder = function (that) {
+        var blockUis = that.reorderer.container.children();
+
+        // used to retrieve the class name for each block
+        // within the blockManager's managedViewComponentRegistry
+        var managedClassNamePattern = sjrk.storyTelling.ui.storyEditor.getManagedClassNamePattern(that.blockManager.options.selectors.managedViewComponents);
```
amb26, line 320: This code has changed a lot since the first review but I still think the "managedClassNamePattern" strategy is quite overcomplicated compared with the requirement - did the alternative strategies, e.g. via a "viewMapper" component or similar get written up as a JIRA somewhere?

src/ui/ui-storyEditor.js
amb26: @cindyli 's code should be identical in all observable effects - I'd be interested in reason for the discrepancy once you find it.
Note that the expression can be simplified yet further:
````
block.applier.change("", 
        order: i,
        firstInOrder: i === 0,
        lastInOrder: i === blockUis.length - 1
});
````
(@cindyli 's version had a few stray semicolons but I imagine you removed them)
(Edited further to compensate for effect of the other typo where "path", "value" where framed at top level)

BlueSlug: Removing the `path` and `value` references did the trick. Thanks @cindyli and @amb26 :)

src/ui/ui-storyEditor.js
BlueSlug: It has now! I've just filed [SJRK-371](https://issues.fluidproject.org/browse/SJRK-371) and have left a comment in the code pointing to it. Many thanks again for your feedback, hopefully we can get to that soon.

src/ui/ui-storyEditor.js
BlueSlug: > > > When dragging the mouse to select some buttons, the "play" button from the text-to-speech shows up. Clicking on it doesn't read anything. If this is a problem, it probably should be addressed in a separate JIRA.
> > > Screen Shot 2020-06-23 at 9 20 53 PM
> > 
> > 
> > We still haven't patched in the fix for FLUID-6512, which would eliminate the page reload when `Play` is clicked, but I'm not sure it's a problem that an empty selection isn't readying anything. The Orator is not designed to read anything other than what is displayed on the screen, if I recally correctly, so no alt-text or aria things. @jobara may be able to confirm this behaviour.
> 
> The Orator only reads the visible text. I suppose you should be able to configure that behaviour although it would likely require some new functions be implemented to know how to parse the text.
> 
> The play button shows up with a selection, if you are finding that it shows up even when no text is selected, perhaps a JIRA against the Orator is warranted.

This has been captured in [FLUID-6523](https://issues.fluidproject.org/browse/FLUID-6523)

BlueSlug:
> > > When dragging the mouse to select some buttons, the "play" button from the text-to-speech shows up. Clicking on it doesn't read anything. If this is a problem, it probably should be addressed in a separate JIRA.
> > > Screen Shot 2020-06-23 at 9 20 53 PM
> > 
> > 
> > We still haven't patched in the fix for FLUID-6512, which would eliminate the page reload when `Play` is clicked, but I'm not sure it's a problem that an empty selection isn't readying anything. The Orator is not designed to read anything other than what is displayed on the screen, if I recally correctly, so no alt-text or aria things. @jobara may be able to confirm this behaviour.
> 
> The Orator only reads the visible text. I suppose you should be able to configure that behaviour although it would likely require some new functions be implemented to know how to parse the text.
> 
> The play button shows up with a selection, if you are finding that it shows up even when no text is selected, perhaps a JIRA against the Orator is warranted.

This has been captured in [FLUID-6523](https://issues.fluidproject.org/browse/FLUID-6523)

BlueSlug:
Ready for review

PR #85, opened 2020-06-15
SJRK-362 Only run deploy on origin repository

gtirloni:
If deploy job runs on forked repository, it'll fail (as expected and desired). This limits where/when the deploy job will run to only the origin repository for SJRK.

### DIFF
--- a/.github/workflows/docker-image-push.yml
+++ b/.github/workflows/docker-image-push.yml
@@ -15,8 +15,9 @@ env:
 
 jobs:
   push:
-    runs-on: ubuntu-latest
     if: github.repository == 'fluid-project/sjrk-story-telling'
+
+    runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/stack_master.yml
+++ b/.github/workflows/stack_master.yml
@@ -11,6 +11,8 @@ env:
 
 jobs:
   deploy:
+    if: github.repository == 'fluidproject/sjrk-story-telling'
+
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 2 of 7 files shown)


gtirloni:
I don't know why but I can't add @BlueSlug as a reviewer. CC'ing here and would appreciate a review. Thanks!

.github/workflows/docker-image-push.yml
```diff
@@ -15,9 +15,10 @@ env:
 
 jobs:
   push:
-    runs-on: ubuntu-latest
     if: github.repository == 'fluid-project/sjrk-story-telling'
 
+    runs-on: ubuntu-latest
```
cindyli, line 20: Right now, all definitions for all branches are in the master branch, which will then be populated to all other branches when the master gets merged into them. I wonder if each branch should only have its own Github Action workflow rather than having the same copy for all branches across all branches?

gtirloni: They are "dormant" in the other branches but it would certainly be possible to clean things up. Would you like me to do that in this PR?

cindyli: Does "dormant" mean the push to individual branch won't trigger the workflow but the push to the master branch will trigger the deployment of all branches? For the clean-up, are you thinking to have the push to branch to trigger the deploy of that branch? 

If so, please go for it. I believe you'll need to send pull request for each branch as not all branches merge the latest master.

gtirloni: Yes, I'm thinking those extra files aren't triggering jobs. Say, a push to the AIHEC branch only triggers the jobs that have `branch: ["aihec..."]`. The jobs for the master branch only are not triggered.

OK, I'll remove the jobs from the branches from the master branch and submit PR's to the branches to add their respective jobs.

gtirloni:
Merging this PR and I'll submit separate PR's to, as requested:

1) Remove extra jobs from the master branch
2) Move those jobs to their respective branches

PR #87, opened 2020-07-03
SJRK-375: Add daily cleanup workflow for staging-stories.floeproject.org

gtirloni:
https://issues.fluidproject.org/browse/SJRK-375

I really hate to have the "Redeploy stack" step here because it's not DRY (the code is already defined in the stack_master.yml workflow). But to trigger another workflow from this one would require 3x more code, the creation of a read-write GitHub token, etc. It's just not worth it for a single step.

### DIFF
--- a/.github/workflows/stack_master.yml
+++ b/.github/workflows/stack_master.yml
@@ -16,6 +16,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+
+      # This step also exists in stack_master_cleanup.yml
       - name: Deploy stack
         uses: appleboy/ssh-action@master
         with:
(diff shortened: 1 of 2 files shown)


gtirloni:
CC @BlueSlug @cindyli

BlueSlug:
Looks good to me! I'm not concerned with the WETness of this workflow, as long as we're aware of it.

Maybe add a comment indicating where to find the "original" version of that step, in case it changes, and a comment in `stack_master.yml` pointing to here?

gtirloni:
> Looks good to me! I'm not concerned with the WETness of this workflow, as long as we're aware of it.
> 
> Maybe add a comment indicating where to find the "original" version of that step, in case it changes, and a comment in `stack_master.yml` pointing to here?

Good point, done!

PR #89, opened 2020-07-14
SJRK-374: solved issue of empty video element being focusable in Firefox

BlueSlug:


themes/base/css/storyTelling.css
```diff
@@ -381,11 +381,16 @@ img.sjrk-st-block-media-preview:not([src]) {
...
-video.sjrk-st-block-media-preview[src=""],
-video.sjrk-st-block-media-preview:not([src]) {
+video.sjrk-st-block-media-preview[src=""] + .sjrk-st-block-video-preview-placeholder,
+video.sjrk-st-block-media-preview:not([src]) + .sjrk-st-block-video-preview-placeholder {
     background: url(../img/icons/Video-background.svg) center center no-repeat;
 }
 
+video:not([controls]) {
+  display: none;
+  height: 0;
```
jobara: Is `height` required here? `display: none;` should remove the element visually, layout spacing, and from the a11y tree.

BlueSlug: This is a good point. I'd left it in there as this is what Normalize does with `audio` elements, but since this rule isn't even present in Normalize after 7.x, I've removed the `height` specification and added an `audio:not([controls])` selector to this rule, to ensure it continues to work as expected if and when we update Normalize

PR #92, opened 2020-07-29
SJRK-122: Run tests with testem and add to CI

jobara:
https://issues.fluidproject.org/browse/SJRK-122


jobara:

tests/ui/testem.js
```diff
@@ -0,0 +1,42 @@
...
+        ]
+    },
+    sourceDirs: {
+        src: "src"
+    },
+    contentDirs: {
+        tests: "tests"
+    },
+    testemOptions: {
+        launch: "Headless Chrome,Headless Firefox",
```
cindyli, line 35: Browser tests run in Chrome and Firefox but not Safari and IE. Combining with the removal of `http://localhost:8081/tests/all-tests.html` request handler, is there a suggested way to run automated browser tests in non-chrome-and-firefox browsers?

jobara: @cindyli testem doesn't have a runner for MS Edge (but it's basically Chrome now anyways), I don't believe we support IE, and there is an [issue](https://github.com/testem/testem/issues/1387) with the latest versions of Safari working with testem. 

Currently the way to run automated tests outside of testem is that you have to provide your own web server to serve the project and then navigate to tests/ui/all-tests.html. This is mentioned in the [README.md](https://github.com/jobara/sjrk-story-telling/blob/SJRK-122/README.md#browser-tests).

cindyli: Running `tests/ui/all-tests.html` with my web server hangs from the first test with the error `jQuery is not defined`.

jobara: @cindyli I was not able to reproduce the issue. You may need to run `npm install` before running the tests. Please let me know if you still have the same issue.

PR #98, opened 2020-11-12
SJRK-425: allow stories without the published flag to be accessed/listed

BlueSlug:
At some point, we should add the `published` flag to the records that don't have it and then revert this work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "sjrk-story-telling",
-    "version": "0.4.0",
+    "version": "0.4.1",
     "description": "The Storytelling Project with Kettle-based server and CouchDB interface",
     "main": "index.js",
     "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "sjrk-story-telling",
-    "version": "0.4.1",
+    "version": "0.4.0",
     "description": "The Storytelling Project with Kettle-based server and CouchDB interface",
     "main": "index.js",
     "scripts": {
--- a/src/server/requestHandlers.js
+++ b/src/server/requestHandlers.js
@@ -117,7 +117,8 @@ sjrk.storyTelling.server.handleGetStory = function (request, dataSource) {
     var noAccessErrorMessage = "An error occurred while retrieving the requested story";
 
     promise.then(function (response) {
-        if (response.published) {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "sjrk-story-telling",
-    "version": "0.4.0",
+    "version": "0.4.1",
```
jobara: We didn't actually tag this one yet, so we don't actually need to do the version bump, but I'm not sure what the practice was before.

BlueSlug: There wasn't a practice before, in fact haha, but I bumped it to keep with semver. Should I drop it back down to 0.4.0?

src/server/requestHandlers.js
```diff
@@ -117,7 +117,8 @@ sjrk.storyTelling.server.handleGetStory = function (request, dataSource) {
     var noAccessErrorMessage = "An error occurred while retrieving the requested story";
 
     promise.then(function (response) {
-        if (response.published) {
+        // only returns published stories or stories that predate the "published" flag
+        if (typeof response.published === "undefined" || response.published) {
```
jobara, line 125: I do worry a bit about defaulting to published. If there are any issues with storing the story, it may get published by accident.

BlueSlug: I agree completely!

This solution was meant to get the stories back in action long enough to give us time to add that value to all the older stories, then revert back to the full restriction.

package.json
BlueSlug: I suppose since we're pre-1.0.0 it doesn't affect things much, either way

src/server/requestHandlers.js
cindyli: I agree the proper solution is to migrate old records in the database with the new field of `"published": true` to avoid adding this kind of special handling which opens a potential hole in the code logic.

However, in case this temporary solution is required for the pressing release, it would be nice to add a TODO comment with the JIRA number to explain the plan. You may also wanna keep SJRK-425 open until the proper solution is implemented.

package.json
BlueSlug: I've set it back to 0.4.0 :)

BlueSlug:
I have filed SJRK-426 to add tests for the database setup functions: https://issues.fluidproject.org/browse/SJRK-426

PR #99, opened 2020-11-13
SJRK-425: restore published story restriction, adjust tests and add migration scripts

BlueSlug:
This PR restores the previous server/db behaviour of restricting public access to published stories only, and it adds `fluid-couch-config` scripts to "migrate" the data of the production sites from `0.3.0` to `0.4.0`.

Broadly, each script contains the entire data set of its respective site, except it's been massaged to be compliant with the server-side autosave functionality introduced in SJRK-289.

To run the scripts, you must connect to each site's web host container and be situated in the project root directory. From there, run the following commands (one for each site):
| Site name | Command |
| ------------- |-------------|
| [AIHEC](https://aihec.inclusivedesign.ca/storyBrowse.html) | `node .\src\server\db\migrations\2020-11-13_0.3.0_to_0.4.0\aihec_stories_0.3.0_to_0.4.0_migration.js` |
| [Inclusive Cities](https://stories.cities.inclusivedesign.ca/storyBrowse.html) | `node .\src\server\db\migrations\2020-11-13_0.3.0_to_0.4.0\cities_stories_0.3.0_to_0.4.0_migration.js` |
| [FLOE Project](https://stories.floeproject.org/storyBrowse.html) | `node .\src\server\db\migrations\2020-11-13_0.3.0_to_0.4.0\floe_stories_0.3.0_to_0.4.0_migration.js` |
| [Karisma](https://karisma-stories.floeproject.org/storyBrowse.html) | `node .\src\server\db\migrations\2020-11-13_0.3.0_to_0.4.0\karisma_stories_0.3.0_to_0.4.0_migration.js` |
| [SJRK](https://stories.sojustrepairit.org/storyBrowse.html) | `node .\src\server\db\migrations\2020-11-13_0.3.0_to_0.4.0\sojustrepairit_stories_0.3.0_to_0.4.0_migration.js` |

### DIFF
--- a/src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/aihec_stories_0.3.0_to_0.4.0_documents.js
+++ b/src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/aihec_stories_0.3.0_to_0.4.0_documents.js
@@ -1,0 +1,305 @@
+/*
+For copyright information, see the AUTHORS.md file in the docs directory of this distribution and at
+https://github.com/fluid-project/sjrk-story-telling/blob/main/docs/AUTHORS.md
+
+Licensed under the New BSD license. You may not use this file except in compliance with this licence.
+You may obtain a copy of the BSD License at
+https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.txt
+*/
+
+// These eslint directives prevent the linter from complaining about the use of
+// globals or arguments that will be prevent in CouchDB design doc functions
+// such as views or validate_doc_update
+
+"use strict";
+
+require("infusion");
+require("kettle");
+
+require("fluid-couch-config");
+
+// sets up the Storytelling Tool database migration using fluid-couch-config
+fluid.defaults("sjrk.storyTelling.server.storiesDb", {
+    gradeNames: ["fluid.couchConfig.pipeline.retrying"],
+    couchOptions: {
+        dbName: "stories"
+    },
+    listeners: {
+        "onSuccess.logSuccess": "fluid.log(SUCCESS)",
+        "onError.logError": "fluid.log({arguments}.0.message)"
+    },
+    dbDocuments: {
+        "8ddcf990-8e18-11ea-820b-c9772cf8ea15":
+        {
+            "type": "story",
+            "value":
+            {
+                "id": "8ddcf990-8e18-11ea-820b-c9772cf8ea15",
+                "published": true,
+                "timestampCreated": "2020-11-13T00:00:00.000Z",
+                "timestampPublished": "2020-11-13T00:00:00.000Z",
+                "title":"Taking Care of Others",
+                "content":
+                [
+                    {
+                        "id":null,
+                        "language":null,
+                        "heading":null,
+                        "blockType":"text",
+                        "text":"When the people were still getting around in wagons, there was a family who needed food before a snowstorm. The husband told his wife he would make a trip to the store to get food. The husband told his wife and children to stay home and she agreed. The husband left. "
+                    },
+                    {
+                        "id":null,
+                        "language":null,
+                        "heading":null,
+                        "blockType":"image",
+                        "mediaUrl":"./uploads/8dd29950-8e18-11ea-820b-c9772cf8ea15.jpg",
+                        "description":null,
+                        "fileDetails":
+                        {
+                            "lastModified":1588604206480,
+                            "lastModifiedDate":"2020-05-04T14:56:46.480Z",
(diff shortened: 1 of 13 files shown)


jobara:
We should take this opportunity now to add the version to the documents in the database.

cindyli:
This migration plan is quite different from what I've envisioned. A couple of questions and concerns:

1. Does each `documents.js` file contain all production data that need to be migrated in that db instance? Where did you get the data from? Did you ask @gtirloni to export them from production databases? When and how to generate these `document.js` files is unclear by looking at this PR. Moreover, I understand SJRK data might be less sensitive as they are accessible via public websites. But it might be improper to expose all these data in a public repo. 

2. All `migration.js` files are identical with one line difference of importing a different `documents.js` file. This duplication should be avoided.

The migration script I envisioned was one single node.js script that accepts the couchDB url as an argument. This script fetches all records from the database, loop thru to update "published" and "mediaUrl", then use CouchDB `/_bulk_docs` to POST updated records to the db.

In terms of running this script, from the GPII experience, the bulk update using `/_bulk_docs` with large number of records may cause CouchDB become irresponsible for a while when CouchDB memory was exhausted. The solution was to have the migration script accept an extra argument of number of records to be updated in one bulk update. But, from @gtirloni's check, the max number of records in production databases is 58. Adding the handling of number of records in a batch doesn't seem necessary for upgrading our databases but might be useful for outside users who has forked and are using this project.

Please create a local test environment with a couple of hundred records to test the migration script before handing it to @gtirloni. Thanks.

BlueSlug:
@cindyli you raise some valid concerns, all of which I share! I was hoping to make only those changes necessary to correct the data, but after some quick testing using `fluid-couch-config`, I saw that in order to update a single field in a single element on the story content array I needed to provide the entire array. I.e. the operation is destructive in that it overwrites previous records with the same ID, hence why the entire data set is included in the script. As I mentioned in IRC, I'm also not comfortable with including this data in a public repo and carrying it around in our version history, but as my goal was to get the stories back in action as soon as possible, I did what I thought would get us back in action quickest. The alternative was to learn the [GPII Universal data migration code](https://github.com/GPII/universal/tree/master/scripts/migration/schema-0.2-GPII-4014) that I wasn't familiar with, and I didn't know how long this would have taken.

If we feel that getting things working again isn't as time-sensitive as I'd previously assumed, and given your concerns, then maybe we should just close this PR and find a more robust and automated solution. That may take less time than trying to massage this one into something passable.

To answer your questions under point 1):
- yes, these scripts contain all production data from each instance,
- I got the data from the live site via the "/stories/" and "/stories/:id" endpoints, which allowed me to get the entire list of stories and the individual story data, respectively (see [this example](https://stories.sojustrepairit.org/stories/23a61a00-07d8-11ea-a5cf-99a2338da816) for story data). This information was reachable once my initial SJRK-425 PR (#98) was deployed.
- After I got the data, I used some tools to help me review and format the data, after which I made the corrections by hand (again, this was for the sake of expedience and I thought this would be faster than writing a tested automated solution)

I feel it's worth mentioning that I did test each of the five scripts in my local environment, starting from both an empty DB and from one with pre-existing data, but not with a couple hundred records. I can do this if we feel it's still necessary.

colinbdclark:
> As I mentioned in IRC, I'm also not comfortable with including this data in a public repo and carrying it around in our version history, but as my goal was to get the stories back in action as soon as possible, I did what I thought would get us back in action quickest. The alternative was to learn the [GPII Universal data migration code](https://github.com/GPII/universal/tree/master/scripts/migration/schema-0.2-GPII-4014) that I wasn't familiar with, and I didn't know how long this would have taken.

Another very good option is to ask for help from a developer who has experience with these kinds of migrations (e.g. @jobara, @cindyli, or @klown), with whom you could work to come up with a solution that addresses these issues that you weren't comfortable with in your solution.

BlueSlug:
> > As I mentioned in IRC, I'm also not comfortable with including this data in a public repo and carrying it around in our version history, but as my goal was to get the stories back in action as soon as possible, I did what I thought would get us back in action quickest. The alternative was to learn the [GPII Universal data migration code](https://github.com/GPII/universal/tree/master/scripts/migration/schema-0.2-GPII-4014) that I wasn't familiar with, and I didn't know how long this would have taken.
> 
> Another very good option is to ask for help from a developer who has experience with these kinds of migrations (e.g. @jobara, @cindyli, or @klown), with whom you could work to come up with a solution that addresses these issues that you weren't comfortable with in your solution.

Yes of course! @cindyli and @klown were providing some guidance and suggested using the GPII solution I mentioned, and @jobara was off that day unfortunately. I had asked about whether the script should live with the repo, but it's clear now that what I had in mind when I asked that question was not what others understood me to mean.

PR #102, opened 2020-11-23
SJRK-425: add DB migration script for 0.3.0 to 0.4.0

BlueSlug:
This pull request adds an automated database migration script to migrate the data from compatibility with version `0.3.0` to `0.4.0`. 

The migration script will:
- retrieve all story data
- make the required updates to a deep copy of the data (see the script proper for more details)
- write the updates to the database
- retrieve all story data again
- verify that all the required changes have been made

The new file `src/server/db/migrations/utils/dbRequestUtils.js` is grifted verbatim from [GPII-universal](https://github.com/GPII/universal/blob/64e97c2c19473461f6e92b5b750c223f8a21a347/scripts/shared/dbRequestUtils.js
), and retains links to the original source and license files of that project.

This PR also adds code to ensure all new stories added to the DB going forward have their `schemaVersion` value set. This value is being tracked separately from the story data, as the client has no need to know the information, nor do we want the client to have any control over this value as it is used for migration purposes. As mentioned in the commit message for 3c45e03, the method we're using to include this value in the database is not preferred, but all other methods have thus far failed.


BlueSlug:
I'm currently working on something to reduce the code duplication in the verification step (where it checks for required values in each story and block)

src/server/dataSource.js
```diff
@@ -61,7 +66,7 @@ fluid.defaults("sjrk.storyTelling.server.dataSource.couch.view", {
 
 // A CouchDB DataSource for a single story, for use on the storyView and storyEdit pages
 fluid.defaults("sjrk.storyTelling.server.dataSource.couch.story", {
-    gradeNames: ["sjrk.storyTelling.server.dataSource.couch.core", "kettle.dataSource.CouchDB"],
+    gradeNames: ["fluid.modelComponent", "sjrk.storyTelling.server.dataSource.couch.core", "kettle.dataSource.CouchDB"],
```
amb26: Doesn't seem to be any need for this to be a modelComponent

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
```diff
@@ -0,0 +1,414 @@
...
+                    aDoc.value.id = aDoc.value.id || aDoc._id; // 2.b
+                    aDoc.value.published = aDoc.value.published || true; // 2.c
+                    aDoc.value.timestampCreated = aDoc.value.timestampCreated || new Date().toDateString(); // 2.d
+                    aDoc.value.timestampPublished = aDoc.value.timestampPublished || new Date().toDateString(); // 2.e
+
+                    var blocks = fluid.copy(aDoc.value.content);
+
+                    fluid.each(blocks, function (block, i) {
+                        block.order = block.order || i; // 2.f
+                        block.firstInOrder = block.firstInOrder || i === 0 ? true : false; // 2.g
```
amb26: For these two lines, recall that ternary booleans are always redundant

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
```diff
@@ -0,0 +1,414 @@
...
+ * @param {Object} options - Where to store the to-be-updated documents
+ *
+ * @return {Array} - The updated documents in the new data structure with the new schema version.
+ */
+sjrk.storyUpdate.updateStoriesData = function (responseString, options) {
+    var allDocs = JSON.parse(responseString);
+    var updatedDocs = [];
+    options.totalNumOfDocs = allDocs.total_rows;
+
+    if (allDocs.rows) {
```
amb26: This test is unnecessary, fluid.each is a no-op on missing arguments. However the intention is clearer if this is a fluid.transform of allDocs.rows (which will also return non-values unchanged)

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
```diff
@@ -0,0 +1,414 @@
...
+                }
+
+                var requiredStoryValues = [
+                    "published",
+                    "timestampCreated",
+                    "timestampPublished"
+                ];
+
+                fluid.each(requiredStoryValues, function (requiredValue) {
+                    if (typeof aDoc.value[requiredValue] === "undefined") {
```
amb26: typeof check is unnecessary (this is only required when using global names which may not have been declared) - aDoc.value[requiredValue] === undefined is fine

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
```diff
@@ -0,0 +1,414 @@
...
+                    fluid.log("Updating the story with ID: ", aDoc._id);
+                    updatedDocs.push(aDoc);
+                }
+            }
+        });
+
+        options.updatedDocs = updatedDocs;
+    }
+
+    return updatedDocs;
```
amb26: Given the return values of all these functions are ignored by the caller, it seems odd to return them.
You could either axe the returns, or else turn this into a slightly more "functional" pipeline involving fluid.promise.fireTransformEvent - this seems to make sense since the yielded value from sjrk.storyUpdate.updateStoriesData is actually supplied to sjrk.storyUpdate.updateDB - the latter can then get the signature sjrk.storyUpdate.updateDB(updatedDocs, options). "responseString" then becomes the previous payload in the chain.
This would also result in some simplification of the logic in gpii.dbRequest.createResponseHandler which accepts a function returning a promise, and then accepts another promise - it could then simply return a promise for its yielded result.
This might be a too ambitious refactoring for the current situation but we should leave the recommendation as a note in case this code gets revisited (I see it has also hopped across from another project once).

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
```diff
@@ -0,0 +1,414 @@
...
+
+                fluid.each(aDoc.value.content, function (block, i) {
+                    var requiredBlockValues = [
+                        "order",
+                        "firstInOrder",
+                        "lastInOrder"
+                    ];
+
+                    fluid.each(requiredBlockValues, function (requiredValue) {
+                        if (typeof block[requiredValue] === "undefined") {
```
amb26: As above

src/server/dataSource.js
```diff
@@ -13,6 +13,11 @@ var fluid = require("infusion");
 require("kettle");
 
 var sjrk = fluid.registerNamespace("sjrk");
+fluid.registerNamespace("sjrk.storyTelling.server.dataSource.couch.story");
+
+// This is the data schema version for the project.
+// All new stories will have their version set to this value.
+sjrk.storyTelling.server.dataSource.couch.story.schemaVersion = "0.4.0";
```
cindyli: I feel it probably makes more sense to separate the version of the data model from the version of the SJRK tool since not every SJRK release will include a data model update.

It's also odd to see the schema version bumps from undefined to 0.4.0.

@BlueSlug, when you have time, It would be nice to have a data model documentation with data structure for every schema version including undefined.

src/server/dataSource.js
BlueSlug: This was leftover from a previous attempt to communicate the schemaVersion to the data record. Removing it!

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
BlueSlug: An old habit from my C# days :) Thanks for pointing that one out!

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
```diff
@@ -0,0 +1,414 @@
...
+ * @param {Array} processArgv - The command line arguments.
+ *
+ * @return {Object} - The options.
+ */
+sjrk.storyUpdate.initOptions = function (processArgv) {
+    var options = {};
+    options.couchDbUrl = processArgv[2];
+    options.maxDocsInBatchPerRequest = Number(processArgv[3]) || sjrk.storyUpdate.defaultMaxDocsInBatchPerRequest;
+
+    options.storiesUrl = options.couchDbUrl + "/_all_docs?include_docs=true&limit=" + options.maxDocsInBatchPerRequest;
```
cindyli: The script is not handling the `maxDocsInBatchPerRequest` properly by calling the recursive function `gpii.dbRequest.processRecursive()` to update documents in batch until all are updated.

Depending on number of documents you will update, as well as if there are other customers hosting their own SJRK servers that will use this migration script and their data volumes, you may either use your current logic but remove the support of `maxDocsInBatchPerRequest`, or add the recursive calls to properly support this feature.

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
```diff
@@ -0,0 +1,414 @@
...
+        fluid.each(allDocs.rows, function (aRow) {
+            var aDoc = aRow.doc;
+            // To filter out the "_design/views" doc that doesn't have the "schemaVersion" field
+            if (aDoc.type === "story") {
+
+                // assume all stories were successfully migrated
+                // if an error is discovered, the cause will be tracked
+                var invalidStoryReasons = [];
+
+                if (!aDoc.schemaVersion || aDoc.schemaVersion < sjrk.storyUpdate.newSchemaVersion) {
```
cindyli: If all docs should have the new schema version, shall the 2nd condition be `aDoc.schemaVersion !== sjrk.storyUpdate.newSchemaVersion`?

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
```diff
@@ -0,0 +1,414 @@
...
+                    invalidStoryReasons.push("Story schema version is not up to date");
+                }
+
+                if (!aDoc.value.id || aDoc.value.id !== aDoc._id) {
+                    invalidStoryReasons.push("Story ID and internal database ID do not match");
+                }
+
+                var requiredStoryValues = [
+                    "published",
```
cindyli: I wonder rather than checking all non-existent keys now exist, shall their values be checked as well? For example:

1. published === true;
2. three order values are set to correct values.

src/server/dataSource.js
BlueSlug: This is a fair point. I had also considered that we should keep the schema version and the tool version separate, and I think I mentioned it during our chat yesterday, though I'm not sure what we should be starting at.

I suppose, since it's an arbitrary number and it doesn't (and in some ways shouldn't) conform to semantic versioning rules, we could start with "1.0", or we could equally use "0.1" since this is a work in progress. Do you have any preference or suggestion? I would want to rename the migration script accordingly, to something like `migrate_unversioned_to_0.1.js`

For documentation of the data model versions, `undefined` consists of all previous states/versions, so it's a little challenging and potentially confusing to list everything. Would it be reasonable to document it as the fields present in v0.3.0 of the tool?

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
BlueSlug: In the interest of saving time, I will opt for removing that feature for this iteration. If we run into issues with the batch size being too large, then we can revisit :)

src/server/dataSource.js
cindyli: I'm fine with either "1.0.0" or "0.0.1", whichever makes more sense to you and @jobara. I would suggest we follow the convention and standard to use 3 numbers in version representing "MAJOR.MINOR.PATCH".

Renaming the migration folder and script to "migrate_unversioned_to_0.0.1" looks great too.

Your point of `undefined` consisting of all previous states/versions makes total sense. I agree it's reasonable to document as of what's in v0.3.0 of the tool.

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
cindyli: Sure. I have no objection.

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
BlueSlug: for the `published` flag, there may be unpublished stories in the future, so I opted to keep it a bit more general. In this case, since story data is not expected to have that value beforehand, checking that value is reasonable.

I'll add checking for the order values, since those are predictable.

cindyli: Well, @BlueSlug, i wonder how future is "in the future"? If the first part of the migration script is looping thru every single doc to set the published flag to `true`, shouldn't the 2nd verification step to make sure this action has happened correctly to all docs?

If at the point this migration script is executed, there aren't any unpublished stories, I think we need to make sure at that point of time, all updates are correct.

My point is migration scripts that run and only run once at a certain point of time are different from the regular tool scripts that consider the future support.

BlueSlug: That's a good point! I'll check the value in this case as well, then.

src/server/db/migrations/2020-11-13_0.3.0_to_0.4.0/migrate_0.3.0_to_0.4.0.js
BlueSlug: This is an excellent idea. I have captured your feedback in [SJRK-429](https://issues.fluidproject.org/browse/SJRK-429) and have added a note above that line in the code.

src/server/db/migrations/schema_0.0.1.md
```diff
@@ -0,0 +1,117 @@
+# Data Schema version: `0.0.1`
```
cindyli: As we talked just now, this data schema documentation fits better into the top "/docs" directory. Please create a README in this migration directory to describe:

1. the goal of this migration;
2. what is changed. Some of your beautiful code comment in the migration script could go in this section;
2. Migration plan (how to perform the migration) including migration steps, what to prepare for each step; what output to expect for each step.

src/server/db/migrations/schema_undefined.md
```diff
@@ -0,0 +1,110 @@
+# Data Schema version: undefined
```
cindyli: Same as above, this doc belongs to the top "/docs" directory. If someone looks for what has been changed from one schema to another, it will be difficult to read and compare each individual schema documentation.

You probably could consider to have one data schema documentation that 1. describe all data schemas; 2. describe changes from one schema to the next. A method I tried before was to use diagram to highlight changes between schemas: https://github.com/fluid-project/preferencesServer/blob/main/doc/dataModel.md

cindyli:
The migration code looks good to me. My only comment is to add a migration plan to describe how to run the migration.

In terms of the comment about creating a data schema docs at the "/docs" directory, @BlueSlug, if you don't have time to address it before the migration, please create a JIRA to address it separately.

BlueSlug:
Ready for review. @gtirloni I have written some instructions on how to use the migration script I prepared, could you please take a look at them and let me know if you have any questions or concerns? :)

The file in question is at `src\server\db\migrations\2020-11-13_undefined_to_0.0.1\README.md`

gtirloni (approved):

cindyli:
@BlueSlug thanks for the super cool and neat documentation. 

One minor problem is, if you check [this section in the migration readme](https://github.com/BlueSlug/sjrk-story-telling/tree/SJRK-425/src/server/db/migrations/2020-11-13_undefined_to_0.0.1#description-of-all-changes), numbers before points include subsections which may not be what you want.

BlueSlug:
> @BlueSlug thanks for the super cool and neat documentation.
> 
> One minor problem is, if you check [this section in the migration readme](https://github.com/BlueSlug/sjrk-story-telling/tree/SJRK-425/src/server/db/migrations/2020-11-13_undefined_to_0.0.1#description-of-all-changes), numbers before points include subsections which may not be what you want.

Oh dear, that's not what they looked like in Atom at all! I wonder why that's happening, I thought my list formatting was correct... Hopefully I can solve it quickly :)

BlueSlug:
> > @BlueSlug thanks for the super cool and neat documentation.
> > One minor problem is, if you check [this section in the migration readme](https://github.com/BlueSlug/sjrk-story-telling/tree/SJRK-425/src/server/db/migrations/2020-11-13_undefined_to_0.0.1#description-of-all-changes), numbers before points include subsections which may not be what you want.
> 
> Oh dear, that's not what they looked like in Atom at all! I wonder why that's happening, I thought my list formatting was correct... Hopefully I can solve it quickly :)

The problem was an incorrect number of spaces for the nested list according to [GitHub-flavoured Markdown rules](https://docs.github.com/en/free-pro-team@latest/github/writing-on-github/basic-writing-and-formatting-syntax#nested-lists) which require the number of spaces in order for the beginning of the nested list items to appear beneath the first character of the higher-level list. Strange and unexpected behaviour, but now I know :)

PR #103, opened 2020-11-25
SJRK-410, SJRK-209: Add editing of published stories

jobara:
Add the ability to load in a published story into the storyEdit page for editing. Also adds "story not found" views for the storyEdit and storyView pages.

https://issues.fluidproject.org/browse/SJRK-410
https://issues.fluidproject.org/browse/SJRK-209


src/ui/base-page-storyEdit.js
```diff
@@ -370,24 +370,35 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+
         try {
-            // localStorage can only store string values
+            /*
+            * load from existing story model if available
+            *
+            * no initalStoryData && no savedStoryData -> create new story
+            * initalStoryData && no savedStoryData -> use initalStoryData
+            * no initialStoryData && savedStoryData -> use savedStoryData
+            * initialStoryData && savedStoryData (different ids) ->use initialStoryData
```
BlueSlug:
```suggestion
            * initialStoryData && savedStoryData (different ids) -> use initialStoryData
```

src/ui/base-page-storyEdit.js
```diff
@@ -370,24 +370,35 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
     };
 
     /**
-     * Loads story content from a given key in the browser's localStorage object.
-     * If a story was successfully loaded, then the current story is updated with
-     * this previously-saved data. If no story is loaded, then a new one is saved
-     * to the server.
+     * Loads an existing story from the server if a story id provide as a query
```
BlueSlug:
```suggestion
     * Loads an existing story from the server if a story id is provided as a query
```

src/ui/base-page-storyEdit.js
```diff
@@ -370,24 +370,35 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+     * paramter. Otherwise loads story content from a given key in the browser's
```
BlueSlug:
```suggestion
     * parameter. Otherwise loads story content from a given key in the browser's
```

src/ui/base-page-storyNotFound.js
```diff
@@ -0,0 +1,60 @@
...
+(function ($, fluid) {
+
+    // The storyNotFound page base grade
+    fluid.defaults("sjrk.storyTelling.base.page.storyNotFound", {
+        gradeNames: ["sjrk.storyTelling.base.page"],
+        storyId: "",
+        events: {
+            onAllUiComponentsReady: {
+                events: {
+                    onViewerReady: "{notFound}.events.onControlsBound"
```
BlueSlug: This event's name should more accurately reflect the composition of this page. Since the `onViewerReady` event in the `storyView` page corresponds to the `storyViewer` component, it would make sense if this event were called `onNotFoundNoticeReady` or something similar, named after the `ui` it's reporting for.

With our slowly-burgeoning codebase increasing in size, it may be good to review these page/ui naming conventions at some point (not for this PR, of course)

src/ui/base-page-storyNotFound.js
```diff
@@ -0,0 +1,60 @@
...
+                }
+            }
+        },
+        components: {
+            notFound: {
```
BlueSlug, line 33: I try to label each component with a comment, something basic like "error message interface" might be appropriate here.

tests/ui/html/base-page-storyNotFound-Tests.html
```diff
@@ -0,0 +1,107 @@
...
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateAware.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateFormControl.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateManager.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateMessage.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateRequestAndRender.js"></script>
+        <script src="../../../node_modules/sinon/pkg/sinon.js"></script>
+
+        <script src="../../../node_modules/fluid-location-bar-relay/src/js/locationBar.js"></script>
+
+        <script src="../../../src/ui/binder.js"></script>
```
BlueSlug: The `binder` component isn't used in this page

tests/ui/html/base-page-storyNotFound-Tests.html
```diff
@@ -0,0 +1,107 @@
...
+        <script src="../../../src/ui/templateManager.js"></script>
+        <script src="../../../src/ui/base-page.js"></script>
+        <script src="../../../src/ui/storyTellingServerUI.js"></script>
+        <script src="../../../src/ui/base-page-storyNotFound.js"></script>
+        <script src="../../../src/ui/story.js"></script>
```
BlueSlug: The `story` grade isn't used in this page

tests/ui/html/base-page-storyNotFound-Tests.html
```diff
@@ -0,0 +1,107 @@
...
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/messageHelper.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/renderer.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateAware.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateFormControl.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateManager.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateMessage.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateRequestAndRender.js"></script>
+        <script src="../../../node_modules/sinon/pkg/sinon.js"></script>
+
+        <script src="../../../node_modules/fluid-location-bar-relay/src/js/locationBar.js"></script>
```
BlueSlug: The `locationBar` component isn't used in this page

tests/ui/html/base-page-storyNotFound-Tests.html
```diff
@@ -0,0 +1,107 @@
...
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/hasRequiredOptions.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/md-client.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/messageHelper.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/renderer.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateAware.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateFormControl.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateManager.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateMessage.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateRequestAndRender.js"></script>
+        <script src="../../../node_modules/sinon/pkg/sinon.js"></script>
```
BlueSlug: Sinon is not required as a dependency on this page

tests/ui/html/base-page-storyNotFound-Tests.html
```diff
@@ -0,0 +1,107 @@
...
+        <link rel="stylesheet" href="../../../themes/base/css/storyTelling.css">
+
+        <script src="../../../node_modules/infusion/dist/infusion-all.js"></script>
+
+        <script src="../../../node_modules/infusion/tests/lib/qunit/js/qunit.js"></script>
+        <script src="../../../node_modules/infusion/tests/test-core/jqUnit/js/jqUnit.js"></script>
+        <script src="../../../node_modules/infusion/tests/test-core/jqUnit/js/jqUnit-browser.js"></script>
+        <script src="../../../node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js"></script>
+
+        <script src="../../../node_modules/fluid-binder/src/js/binder.js"></script>
```
BlueSlug: The `binder` component is not used in this page grade (only the Edit page), so it's not required for the tests

tests/ui/html/base-page-storyNotFound-Tests.html
```diff
@@ -0,0 +1,107 @@
...
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateMessage.js"></script>
+        <script src="../../../node_modules/fluid-handlebars/src/js/client/templateRequestAndRender.js"></script>
+        <script src="../../../node_modules/sinon/pkg/sinon.js"></script>
+
+        <script src="../../../node_modules/fluid-location-bar-relay/src/js/locationBar.js"></script>
+
+        <script src="../../../src/ui/binder.js"></script>
+        <script src="../../../src/ui/templateManager.js"></script>
+        <script src="../../../src/ui/base-page.js"></script>
+        <script src="../../../src/ui/storyTellingServerUI.js"></script>
```
BlueSlug: I believe this file is also unnecessary for these tests

tests/ui/html/base-page-storyNotFound-Tests.html
```diff
@@ -0,0 +1,107 @@
...
+        <script src="../../../src/ui/ui-menu.js"></script>
+
+        <script src="../../../src/ui/UIOMultilingualDemo.js"></script>
+
+        <!-- Required to send coverage data at the end of the test run on this page -->
+        <script src="/testem.js"></script>
+        <script src="/coverage/client/coverageSender.js"></script>
+
+        <script src="../js/base-page-storyNotFoundTests.js"></script>
+        <script src="../js/testUtils.js"></script>
```
BlueSlug: If none of the `testUtils` functions are used, it is not needed on this page

tests/ui/js/base-page-storyEditTests.js
```diff
@@ -1464,6 +1464,185 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
+        }
+    };
+
+    /*
+    * load from existing story model if available
+    *
+    * no initalStoryData && no savedStoryData -> create new story
+    * initalStoryData && no savedStoryData -> use initalStoryData
+    * no initialStoryData && savedStoryData -> use savedStoryData
+    * initialStoryData && savedStoryData (different ids) ->use initialStoryData
```
BlueSlug:
```suggestion
    * initialStoryData && savedStoryData (different ids) -> use initialStoryData
```

tests/ui/js/base-page-storyNotFoundTests.js
```diff
@@ -0,0 +1,121 @@
...
+                sequence: [{
+                    event: "{storyNotFoundTest > sjrk.storyTelling.base.page.testStoryNotFound}.events.onAllUiComponentsReady",
+                    listener: "sjrk.storyTelling.base.page.storyNotFoundTester.assertRendering",
+                    args: ["{storyNotFound}.notFound", "{that}.options.testOpts.expectedMessages"]
+                }]
+            }]
+        }]
+    });
+
+    sjrk.storyTelling.base.page.storyNotFoundTester.collapseText = function (str) {
```
BlueSlug, line 134: Please add JSDoc comments for these two functions

tests/ui/js/storyTellingServerUITests.js
```diff
@@ -80,12 +80,15 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
                 authoringEnabled: true
             }
         },
+        retrievedStory: "{}",
+        // retrievedStory: "{\"title\":\"test story\"}",
```
BlueSlug: Is this line expected to be used in the future?

themes/aihec/js/aihec.js
```diff
@@ -75,6 +75,15 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
         }]
     });
 
+    // Applies the AIHEC shell to the storyNotFound page
+    fluid.defaults("sjrk.storyTelling.aihec.page.storyNotFound", {
+        gradeNames: ["sjrk.storyTelling.aihec.page", "sjrk.storyTelling.base.page.storyNotFound"],
+        distributeOptions: {
```
BlueSlug: If the theme-specific template is identical to the base theme's, it isn't necessary to duplicate it, nor to update the template path. This applies to all themes, of course.

You'll still need to have this theme-specific grade defined, but no options are necessary in this case.

BlueSlug (changes requested):
Looks pretty good! Tests are passing in my environment, and I didn't notice any issues while testing.

One thing to note (which I noted :)) is that since the theme-specific templates for the storyNotFound page aren't any different from the base theme, they aren't actually necessary. The server will automagically fall back to the base theme template when it can't find a specified file in the theme's template directory. The theme-specific grades for each page are still necessary, but they don't need to specify a different template in order to work properly.

src/ui/base-page-storyNotFound.js
jobara: copy and paste error ;( Thanks for catching it.

themes/aihec/js/aihec.js
jobara: Thanks, I hadn't realized that.

src/ui/base-page-storyEdit.js
```diff
@@ -370,24 +370,35 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
...
      *
      * @param {String} storyAutosaveKey - the key to load the story content from
      * @param {Component} storyEdit - an instance of `sjrk.storyTelling.base.page.storyEdit`
      */
     sjrk.storyTelling.base.page.storyEdit.initializeStory = function (storyAutosaveKey, storyEdit) {
+
         try {
-            // localStorage can only store string values
+            /*
+            * load from existing story model if available
```
cindyli, line 386: Thanks for documenting this list of scenarios. It's helpful to understand the situation.

But it looks overwhelming by covering these different combinations to guess what story data should be loaded.

I wonder if this can be simplified by passing in an "edit"/"create" state as a model value. Then the logic would be: 1. at the "create" state, load saved story data from the local storage; 2. at "edit", load data retrieved from the server.

tests/server/testServerWithStorage.js
```diff
@@ -338,9 +304,54 @@ sjrk.storyTelling.server.testServerWithStorageDefs = [{
                     files: {
                         "file": "{testCaseHolder}.options.testUploadOptions.testPNGFile"
                     }
-                },
+                }
+            }
+        }
+    },
+    dynamicComponents: {
+        replaceFile: {
```
cindyli, line 312: One issue I think should have been addressed in @BlueSlug's SJRK-289 pull request, which has already been merged, is when re-uploading a media file, the previous media file should be removed from the the /uploads directory, so the server doesn't end with many stray images. 

However, this issue still exists here. Is there a changed plan to address it?

jobara: We needed to keep the original files because when editing a published story you could delete a published file. If the story being edited isn't yet published, the currently published story would look broken because the file is missing. 

I've filed a new issue, [SRJK-431](https://issues.fluidproject.org/browse/SJRK-431), for automatically cleanup orphaned files.

src/ui/base-page-storyEdit.js
jobara: I'm not sure that's possible. In particular for the case where you are editing a published story and also want to make use of the auto save local storage. In the implementation I've currently included, it will use the local storage if that's available with the same id as the incoming changes from the server.

PR #104, opened 2020-12-07
SJRK-435: fix ToC preview bug

BlueSlug:
UPDATE: This bug is solved by always hiding the UIO Table of Contents on the Edit page

Jira: https://issues.fluidproject.org/browse/SJRK-435

### DIFF
--- a/themes/base/storyEdit.html
+++ b/themes/base/storyEdit.html
@@ -76,6 +76,11 @@
 <link rel="stylesheet" href="/node_modules/infusion/src/framework/core/css/fluid.css">
 
 <script src="src/storyTellingServerUI.js"></script>
+<style>
+    .flc-toc-tocContainer {
+        display: none !important;
+    }
+</style>
 
 </head>
 <body>
--- a/themes/base/storyEdit.html
+++ b/themes/base/storyEdit.html
@@ -76,6 +76,15 @@
 <link rel="stylesheet" href="/node_modules/infusion/src/framework/core/css/fluid.css">
 
 <script src="src/storyTellingServerUI.js"></script>
+
+<!--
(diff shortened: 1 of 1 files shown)


themes/base/storyEdit.html
```diff
@@ -76,6 +76,11 @@
 <link rel="stylesheet" href="/node_modules/infusion/src/framework/core/css/fluid.css">
 
 <script src="src/storyTellingServerUI.js"></script>
+<style>
+    .flc-toc-tocContainer {
+        display: none !important;
+    }
+</style>
```
jobara, line 92: Rather than adding a `<style>` element, it may be cleared just to add an it as an inline style directly on toc container. It would also be good to leave a comment with a link to the JIRA to explain why it's hidden.

jobara:
Rather than revert the commit, could you reset back and only include the change to the styling. It would also be good to update the JIRA with some information about approaches tried and the why the current solution was chosen. Just so we can keep a history of things.

themes/base/storyEdit.html
BlueSlug: I had tried this, since inline styling has the highest specificity, but turning the TOC preference on and off will make the element show again.

BlueSlug:
I've added [this comment](https://issues.fluidproject.org/browse/SJRK-435?focusedCommentId=42134&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-42134) to the Jira, detailing why the particular solution we chose is used and what other options were considered.

Ready for more review :)

jobara:
merged at c52e54e2aed53540cae12718dba3c48e520a0eb4

PR #105, opened 2020-12-08
SJRK-405: Implement server side logic for signup/login/logout and session support

jobara:
https://issues.fluidproject.org/browse/SJRK-405


src/server/dataSource.js
```diff
@@ -64,6 +64,22 @@ fluid.defaults("sjrk.storyTelling.server.dataSource.couch.view", {
     }
 });
 
+// A CouchDB DataSource for retrieving a story by authorID and storyId
+fluid.defaults("sjrk.storyTelling.server.dataSource.couch.authorStoriesView", {
+    gradeNames: ["sjrk.storyTelling.server.dataSource.couch.core"],
+    // TODO: investigate overriding the resolveUrl invoker to all for providing the query paramters via
```
BlueSlug: Could you file a Jira for this and link to it from this comment, please?

src/server/requestHandlers.js
```diff
@@ -131,15 +132,61 @@ sjrk.storyTelling.server.handleGetStory = function (request, dataSource) {
...
 
         request.events.onError.fire({
+            statusCode: 404,
             isError: true,
             message: noAccessErrorMessage
         });
     });
 };
 
+// Kettle request handler for a single story (View)
```
BlueSlug:
```suggestion
// Kettle request handler for getting a single story (Edit)
```

src/server/requestHandlers.js
```diff
@@ -131,15 +132,61 @@ sjrk.storyTelling.server.handleGetStory = function (request, dataSource) {
...
+    invokers: {
+        handleRequest: {
+            funcName: "sjrk.storyTelling.server.handleGetEditStory",
+            args: ["{request}", "{server}.storyByAuthorDataSource"]
+        }
+    }
+});
+
+/**
+ * Gets a single story's data from the database, parses it, rebuilds any URLs to
```
BlueSlug: No URL rebuilding is happening any more as of `0.4.0`, I forgot to remove that part of the comment for `sjrk.storyTelling.server.handleGetStory`. This part of the sentence should be removed for that function as well as this one here

src/server/requestHandlers.js
```diff
@@ -583,3 +660,143 @@ fluid.defaults("sjrk.storyTelling.server.nodeModulesHandler", {
         }
     }
 });
+
+/********************************************
+ * Request Handlers for signup/login/logout *
+ *******************************************/
```
BlueSlug: This comment style (along with its partner at the bottom) doesn't match the rest of the project. I see that it's being used to delineate which grades are handlers for stories & files and which are for session management, and I wonder if it might not make more sense to split these into their own file, e.g. `requestHandlers-session.js` and `requestHandlers-story.js`, or some other more meaningful/practical naming convention. What do you think?

src/server/serverSetup.js
```diff
@@ -41,15 +45,65 @@ fluid.defaults("sjrk.storyTelling.server", {
...
                 distributeOptions: {
-                    record: "@expand:kettle.resolvers.env(COUCHDB_URL)",
-                    target: "{that sjrk.storyTelling.server.dataSource.couch.core}.options.host"
+                    couchDBURL: {
+                        record: "@expand:kettle.resolvers.env(COUCHDB_URL)",
+                        target: "{that sjrk.storyTelling.server.dataSource.couch.core}.options.host"
+                    },
+                    couchDBAuthorsURL: {
+                        record: "@expand:kettle.resolvers.env(COUCHDB_URL)",
+                        target: "{that gexpressUserUtils}.options.dataSourceConfig.host"
```
BlueSlug:
```suggestion
                        target: "{that expressUserUtils}.options.dataSourceConfig.host"
```

src/server/requestHandlers.js
```diff
@@ -583,3 +660,143 @@ fluid.defaults("sjrk.storyTelling.server.nodeModulesHandler", {
...
+            args: ["{request}", "{expressUserUtils}"]
+        }
+    }
+});
+
+/**
+ * Logs in an existing user with the password and username provided in the request.
+ *
+ * @param {Object} request - a Kettle request
+ * @param {fluid.express.user.utils} expressUserUtils - an instance of `fluid.express.user.utils`
```
BlueSlug: In other parts of the codebase, I've put `{Component}` in cases where the arg is an Infusion component. For consistency, I'd suggest either setting this to that or setting all the other ones to the fully-qualified grade name. The latter might be out of scope for this Jira, in which case we should file a new one to look into what we should be doing and then implementing whatever solution we come up with.

src/server/serverSetup.js
```diff
@@ -41,15 +45,65 @@ fluid.defaults("sjrk.storyTelling.server", {
...
+                        source: "{that}.options.session",
+                        target: "{that > kettle.middlewareHolder > session}.options"
+                    }
                 },
                 components: {
+                    // user authentication
+                    expressUserUtils: {
+                        type: "fluid.express.user.utils",
+                        options: {
+                            iterations: 10,
```
BlueSlug: Is 10 iterations enough? I've seen 100, 1,000 or even 10,000 iterations suggested in order to defend against brute-force attacks, especially so with faster hashing algorithms. That said, if we're heading towards cloud hosting, too high a number could make hosting very expensive...

src/server/serverSetup.js
```diff
@@ -252,3 +334,11 @@ sjrk.storyTelling.server.getThemePath = function (theme, themeFolder) {
 
     return themePath;
 };
+
+sjrk.storyTelling.server.makeMemorySessionStore = function () {
+    // TODO: Currently using https://www.npmjs.com/package/memorystore as it is a production ready memory store;
```
BlueSlug, line 356: If this `TODO` is not intended to be tackled in SJRK-405, please file a Jira to do that work and add a link to that Jira here

src/server/serverSetup.js
```diff
@@ -252,3 +334,11 @@ sjrk.storyTelling.server.getThemePath = function (theme, themeFolder) {
 
     return themePath;
 };
+
+sjrk.storyTelling.server.makeMemorySessionStore = function () {
```
BlueSlug: Please add a JSDoc comment for this function

tests/server/dataSource-tests.js
```diff
@@ -0,0 +1,606 @@
...
+        },
+        "author2": {
+            "type": "user",
+            "email": "test2@example.com",
+            "authorID": "author2",
+            "username": "test2@example.com"
+        }
+    },
+    stories: {
+        "story1": {
```
BlueSlug: It may be helpful to name these test cases after their contents or purpose

tests/server/dataSource-tests.js
```diff
@@ -0,0 +1,606 @@
...
+
+/**
+ * A wrapper around `jqUnit.assertDeepEq` except it strips `_rev` out of the actual response as this is generated value.
+ * The remaining properties are compared with `jqUnit.assertDeepEq`.
+ *
+ * @param {String} msg - The assertion message
+ * @param {Object} expected - The expected value to compare against the `actual` value.
+ * @param {Object} actual - The actual value, minus `_rev` property, compared against the `expected` value.
+ */
+sjrk.test.storyTelling.server.dataSource.tester.assertResponse = function (msg, expected, actual) {
```
BlueSlug: A similar filtering out of `_rev` is done in `sjrk.storyTelling.server.testServerWithStorageDefs.verifyStoryPersistence`. Perhaps that piece of code and this function could be combined in a shared function in `serverTestUtils.js`.

tests/server/dataSource-tests.js
```diff
@@ -0,0 +1,606 @@
...
+ * @param {String} msg - The assertion message
+ * @param {Object} actual - The actual error returned
+ * @param {Object} [options] - (optional) values to be interpolated into the error message
+ * @param {String} [options.method] - (optional) the HTTP method to interpolate into error message; default "GET"
+ * @param {String} [options.storyId] - (optional) the storyId to interpolate into the error message; default ""
+ * @param {String} [options.rev] - (optional) the rev to interpolate into the error message
+ */
+sjrk.test.storyTelling.server.dataSource.tester.assertMissingError = function (msg, actual, options) {
+    var tokens = fluid.merge("replace", {}, {method: "GET", storyID: ""}, options || {});
+    var expected = {
```
BlueSlug: Would the expected values be better defined outside of this function and passed in?

BlueSlug (commented):
Looking pretty good!

src/server/dataSource.js
jobara: https://issues.fluidproject.org/browse/SJRK-443

src/server/serverSetup.js
jobara: This is the default from the underlying library actually. We could make it some other value as needed.

src/server/serverSetup.js
jobara: https://issues.fluidproject.org/browse/SJRK-444

src/server/requestHandlers.js
jobara: I've switched it to `{Component}` for consistency. We should review this in the future. In Infusion the practice has moved to use the component grade; however, a description is still required by the JSDoc validation.

tests/server/dataSource-tests.js
```diff
@@ -0,0 +1,565 @@
...
+                        }
+                    }
+                }
+            }
+        },
+        // a DataSource to get a list of stories
+        viewDataSource: {
+            type: "sjrk.storyTelling.server.dataSource.couch.view",
+            options: {
+                host: "http://localhost:6789"
```
BlueSlug: Perhaps a distributeOptions could be used to inject this host configuration into each test dataSource. This is fine either way, though :)

tests/server/dataSource-tests.js
```diff
@@ -0,0 +1,565 @@
...
+});
+
+fluid.defaults("sjrk.test.storyTelling.server.dataSource.tester", {
+    gradeNames: ["fluid.test.testCaseHolder"],
+    members: {
+        rev: null
+    },
+    testOpts: {
+        storiesViewError: {
+            "reason": "ddoc _design/stories has no view named notAView",
```
BlueSlug, line 221: Is this meant to read "ddoc" or "doc"? If it's the former, it may be more clear if it says "design doc"

jobara: that's an error straight from the database, it isn't something I defined.

BlueSlug: Curious! But okay, let's just leave it as-is, in that case

src/server/db/authors-dbConfiguration.js
```diff
@@ -0,0 +1,50 @@
...
+require("kettle");
+
+var sjrk = fluid.registerNamespace("sjrk");
+require("fluid-couch-config");
+
+// sets up the Storytelling Tool database using fluid-couch-config
+fluid.defaults("sjrk.storyTelling.server.authorsDb", {
+    gradeNames: ["fluid.couchConfig.pipeline.retrying"],
+    couchOptions: {
+        dbName: "authors"
```
cindyli: Just curious, is "authors" a separated database that only contains the "user" document type? What's the reason to decide on using a separate database rather than making use of the existing "stories" database with a new "user" document type?

With the current logic, using 2 databases works the same way. But it feels strange using multiple databases to support an app, which is unnecessary.

src/server/db/singleNodeConfiguration.js
```diff
@@ -44,3 +44,11 @@ fluid.defaults("sjrk.storyTelling.server.globalChangesDb", {
         dbName: "_global_changes"
     }
 });
+
+// includes the users database name
+fluid.defaults("sjrk.storyTelling.server.storyUsersDb", {
```
cindyli: What is this "users" database? It doesn't seem to be used anywhere.

jobara: I believe it's couchDB's user database for accounts to itself.

jobara: oh actually this one might be a mistake, might have been leftovers from trying things out early on.

src/server/db/authors-dbConfiguration.js
jobara: That the user accounts are managed through fluid-express-users library, but I think it could actually be in the same database as stories. I'll look into that.

src/server/storyRequestHandlers.js
```diff
@@ -157,28 +202,51 @@ fluid.defaults("sjrk.storyTelling.server.saveStoryHandler", {
...
         });
+
+        return;
     }
+
+    var id = request.req.params.id;
+
+    dataSource.get({directStoryId: id}).then(function (response) {
+        if (request.req.session.authorID === response.authorID) {
+            sjrk.storyTelling.server.saveStoryToDatabase(dataSource, request.req.session.authorID, request.req.body, request.events.onSuccess, request.events.onError);
```
cindyli, line 237: Could you add a code comment here and at line 226 that this line is to update an existing story while line 226 is to create a new story.

cindyli:
It would be helpful to add server API documentation when you have time. Thanks.

src/server/db/authors-dbConfiguration.js
BlueSlug: I looked into this when I first saw @jobara's choice. Turns out using separate databases is actually preferable, on account of the way index building works in CouchDB, though the performance gains to us will be minimal. Here's a very interesting thread on the topic, I recommend reading the answers other than the one the OP chose: https://stackoverflow.com/questions/10503089/couchdb-multiple-databases

jobara: @BlueSlug that's a very interesting point. Unfortunately for me I already combined them. @cindyli and @BlueSlug let me know if you think we should split them back out.

cindyli: It's great to know separating data into multiple databases helps with the CouchDB performance. Thanks for the good thread, @BlueSlug. After reading the thread, I'm fine with either.

If @jobara has already combined it, I would prefer leave as combined. As @BlueSlug said, the performance gain for SJRK will be minimal. Moreover, using one database simplifies the code and is easier for backup and maintenance.

jobara:
@cindyli this PR is ready for review.

src/server/db/story-dbConfiguration.js
```diff
@@ -141,7 +151,7 @@ sjrk.storyTelling.server.storiesDb.storyTagsFunction = function (doc) {
  * @param {Object} doc - a document to evaluate in the view
  */
 sjrk.storyTelling.server.storiesDb.storiesByIdFunction = function (doc) {
-    if (doc.value.published) {
+    if (doc.type === "story" && doc.value.published) {
```
cindyli, line 154: The condition of checking `doc.value.published` shows a more meaning name for this view might be "publishedStoriesById`.

jobara: Makes sense. I've renamed it.

src/server/serverSetup.js
```diff
@@ -41,15 +46,69 @@ fluid.defaults("sjrk.storyTelling.server", {
                     secrets: "@expand:sjrk.storyTelling.server.resolveJSONFile({that}.options.secureConfig.secretsConfigPath)"
                 },
                 port: "{that}.options.globalConfig.port",
+                session: {
+                    store: "@expand:sjrk.storyTelling.server.makeMemorySessionStore({that}.options.globalConfig.sessionCheckPeriod)",
+                    middlewareOptions: {
+                        secret: "{server}.options.secureConfig.secrets.session"
+                    }
+                },
+                account: "{that}.options.secureConfig.secrets.account",
```
cindyli: What is this account for? `secrets.json.example` does not provide an example of this account value.

This affects line 69-72 too which distributes this value.

src/server/serverSetup.js
```diff
@@ -207,6 +279,21 @@ fluid.defaults("sjrk.storyTelling.server.app.storyTellingHandlers", {
             route: "/clientConfig",
             method: "get"
         },
+        signupHandler: {
```
cindyli, line 287: This pull request does not contain the tests and documentation for these new endpoints. Is it planned to be addressed in a separate ticket?

src/server/db/story-dbConfiguration.js
```diff
@@ -170,3 +191,16 @@ sjrk.storyTelling.server.storiesDb.validateFunction = function (newDoc) {
...
+
+/**
+ * This function is used as a "view" design doc once it's migrated to CouchDB.
+ * Will locate user/author accounts.
+ *
+ * @param {Object} doc - a document to evaluate in the view
+ */
+sjrk.storyTelling.server.storiesDb.byUsernameOrEmail = function (doc) {
+    if (doc.type === "user") {
+        emit(doc.username, doc);
```
cindyli: I know nothing about the client side logic. As the server logic in this pull request always returns the email address as a login success signal, I wonder if a user could login via either the username or email address, or just email address. In the case of the latter, it might be better to remove this line in order to reduce the index volume and prevent a potential security issue.

src/server/serverSetup.js
jobara: This provides a means of configuring [`fluid.express.user.utils`](https://github.com/fluid-project/fluid-express-user/blob/e826290beb4fdb416a4fedf602803ac1bcef4f37/src/js/server/utils.js#L28). Not sure how to show that in the `secrets.json.example`. Any suggestions?

src/server/serverSetup.js
jobara: The authentication related handlers are included in [testSeverWithStorage.js](https://github.com/jobara/sjrk-story-telling/blob/SJRK-405/tests/server/testServerWithStorage.js) as they are required for saving/editing stories.

src/server/db/story-dbConfiguration.js
jobara: There are no explicit usernames anymore, only an e-mail address to identify authors. Thanks for spotting this leftover bit. I've removed it, and have also renamed the name of the function. There are some other references to byUsernameOrEmail but these are used internally by fluid-express-user, so I have left them as is.

src/server/serverSetup.js
jobara: They are also tested in [/testServerAuthentication.js](https://github.com/jobara/sjrk-story-telling/blob/SJRK-405/tests/server/testServerAuthentication.js)

src/server/serverSetup.js
cindyli: Can you point me to the doc or code that show how this `account` is used for configuring `fluid.express.user.utils`? I didn't find its usage in the given link. Just want to get an idea about its value type and the logic behind it. Thanks.

jobara: @cindyli the link above is to the components defaults. We add fluid.express.user.utils as a [subcomponent](https://github.com/jobara/sjrk-story-telling/blob/89fe0ec255b7fef24aa12c107a7f3b198ac257df/src/server/serverSetup.js#L76-L102) to the server. These defaults can be set/overridden by passing in configuration to the secrets.json file under the `account` property. These are relayed via the lines you mentioned above. The component options for [account](https://github.com/jobara/sjrk-story-telling/blob/89fe0ec255b7fef24aa12c107a7f3b198ac257df/src/server/serverSetup.js#L55) and the [options distributions](https://github.com/jobara/sjrk-story-telling/blob/89fe0ec255b7fef24aa12c107a7f3b198ac257df/src/server/serverSetup.js#L69-L72) to send this to the instance of fluid.express.user.utils.

jobara:
@cindyli I'm trying to write up some docs on the server endpoints as well. Working through how to format them, but I have an idea now.

jobara:
@cindyli I've added the server endpoints documentation. Ready for more review.

cindyli:
The server endpoints doc looks great. Thanks, @jobara.

src/server/serverSetup.js
cindyli: I think now I understand better of what you said about "provides a means of configuring fluid.express.user.utils". A couple of suggestions:

1. Please consider to use a more meaningful variable name such as `expressUserUtilsConfig`;
2. I'm not sure if it's a good idea to use `secrets.json` to distribute component options. These options only make sense to SJRK developers while `secrets.json` gives a feeling of being configured by website admins who unlikely know about infusion and what component options are;
3. If you decided to use `secrets.json` for this config, please add documentation, for example in README where `secrets.json` is explained. (If `secrets.json` is a json5 file, comments for each value in that file would be easier to access.)

jobara: @cindyli I've added some documentation about configuration, add some examples to the secrets.json.example file, and renamed some things. I think mostly you can just use the kettle config to configure expressUserUtils, but there are some parts that we may want to express in the secrets file, so I've illustrated those in the example.

src/server/serverSetup.js
```diff
@@ -41,15 +46,69 @@ fluid.defaults("sjrk.storyTelling.server", {
...
+                    couchDBAuthorsURL: {
+                        record: "@expand:kettle.resolvers.env(COUCHDB_URL)",
+                        target: "{that expressUserUtils}.options.dataSourceConfig.host"
+                    },
+                    sessionOptions: {
+                        source: "{that}.options.session",
+                        target: "{that > kettle.middlewareHolder > session}.options"
+                    },
+                    authorCredentialConfig: {
+                        source: "{that}.options.account",
```
cindyli: The differentiation of the kettle server config and account credential config in various config files makes a lot of sense. Thanks, @jobara.

With the renaming at line 55, the source reference of `account` should also be renamed to `authorCredentialConfig`. It would be nice to have tests for the proper handling/distribution of secrets.json.

jobara: Thanks for catching the leftover bit. I've addressed that and added some test for the distributions of the secrets.

cindyli:
This pull request looks great to me. 

@BlueSlug, can you help to do a full manual test? Once it passes the test, feel free to merge. Thanks.

docs/CONFIGURATION.md
```diff
@@ -0,0 +1,75 @@
...
+secret and configure how author credentials are encrypted.
+
+For more information about the author credential options, see [crypto](https://nodejs.org/api/crypto.html).
+
+An example, [secrets.json.example](../secrets.json.example),
+has been provided and can be used as a template to create your own file.
+
+```JSON5
+{
+    "adminPass": "admin password",
```
BlueSlug, line 67: The section of `README.md` titled "Running the site" mentions what the contents of `secrets.json` need to be in order to run the site. I suggest either updating that section to reflect the contents here, or linking directly to this file

BlueSlug (approved):
The tests are all passing without issue, now :)

jobara:
I believe @BlueSlug gave his [approval](https://matrix.to/#/!mDigWuPWilYGiEcOCA:matrix.org/$8lcQ77Vz-nE2jIlbd6AQEq1lv3l7XjkG163ta1EeYUg?via=matrix.org) in the channel. I'll self merge this PR.

jobara:
Merged at 3845afd05c22ba96283c4558581b90e69ea47319

PR #106, opened 2021-01-05
SJRK-446 Update to Node.js 14.15.4 (security fixes)

gtirloni:
https://nodejs.org/en/blog/release/v14.15.4/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:14.15.1-alpine
+FROM node:14.15.4-alpine
 
 WORKDIR /usr/src/app
 


gtirloni:
I don't know what's happening with macOS tests getting stuck and I don't have Apple hardware to test on, unfortunately. I've tried to re-run the jobs, and with Node.js 14.x, they fail on macOS at different points.

EDIT: Now the tests are failing on Node.js 12.x and Ubuntu, so it doesn't seem related to a specific platform.

EDIT2: I can reproduce the failure locally with Node.js 14.x on Fedora Linux.

gtirloni:
It seems something in `html/base-page-Tests.html` is making tests randomly fail.

BlueSlug (approved):
I suspect this (and the often-hanging CI jobs in other PRs) is related to [SJRK-372](https://issues.fluidproject.org/browse/SJRK-372), and it smells strongly of a race condition.

FWIW, the tests ran fine in my local environment (Node 14.13.0) and when running the browser tests directly in Chrome 87, Firefox 84 and Edge 87.

I also throttled my sixteen 3.3 GHz threads by 95% to see if slowing things down would trigger the issue, but no luck. Thinking about this some more, if it's related to requesting and loading additional resources after the script begins running, then speeding it up might actually be better at reproducing the issue. Too bad I don't have a Threadripper :P

What does this mean for your PR? Since the issue is previously known and doesn't seem related to your change, and since the tests are otherwise passing without issue, I approve merging this. We should add a link to this PR from the SJRK-372 Jira.

Edit: I added a comment on SJRK-372

jobara:
merged at 659dfd5f649293b92962118c5554705eabe8c2ad

PR #108, opened 2021-01-25
SJRK-372: Address race condition

jobara:
https://issues.fluidproject.org/browse/SJRK-372

### DIFF
--- a/src/ui/base-page.js
+++ b/src/ui/base-page.js
@@ -160,6 +160,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
      *
      * @param {Component} pageComponent - the `sjrk.storyTelling.base.page` that will accept the preferences
      * @param {Component} cookieStore - a fluid.prefs.cookieStore containing the data to laod
+     * @return {Promise} - the promise returned for getting the cookie data.
      */
     sjrk.storyTelling.base.page.getStoredPreferences = function (pageComponent, cookieStore) {
         var promise = cookieStore.get();
@@ -172,6 +173,8 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
         }, function (error) {
             pageComponent.events.onPreferenceLoadFailed.fire(error);
         });
+
+        return promise;
     };
 
     /**
--- a/src/ui/base-page.js
+++ b/src/ui/base-page.js
@@ -56,11 +56,21 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
                 target: "{that ui blockManager templateManager}.options.templateConfig.resourcePrefix"
             },
             "timeBased.stopMediaPlayerOnContextChange": {
-                record: { "{page}.events.onContextChangeRequested": "{that}.stopMediaPlayer" },
+                record: {
+                    "{page}.events.onContextChangeRequested": {
+                        listener: "{sjrk.storyTelling.blockUi.timeBased}.stopMediaPlayer",
+                        namespace: "stopMediaPlayer"
+                    }
+                },
                 target: "{that sjrk.storyTelling.blockUi.timeBased}.options.listeners"
             },
             "ui.requestResourceLoadOnRenderAllUiTemplates": {
-                record: { "{sjrk.storyTelling.base.page}.events.onRenderAllUiTemplates": "{templateManager}.events.onResourceLoadRequested.fire" },
+                record: {
+                    "{sjrk.storyTelling.base.page}.events.onRenderAllUiTemplates": {
+                        listener: "{templateManager}.events.onResourceLoadRequested.fire",
+                        namespace: "requestResourceLoad"
+                    }
+                },
                 target: "{that sjrk.storyTelling.ui}.options.listeners"
             },
             "templateManager.uiLanguageToTemplateManager": {
@@ -98,16 +108,7 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
                 funcName: "sjrk.storyTelling.base.page.getStoredPreferences",
                 args: ["{that}", "{cookieStore}"]
             },
-            "{menu}.events.onInterfaceLanguageChangeRequested": [{
-                func: "{that}.applier.change",
-                args: [["persistedValues", "uiLanguage"], "{arguments}.0.data"],
-                namespace: "changeUiLanguage"
-            },
-            {
-                func: "{that}.events.onContextChangeRequested.fire",
-                namespace: "onContextChangeRequested",
-                priority: "last"
-            }]
+            "onRenderAllUiTemplates.onContextChangeRequested": "{that}.events.onContextChangeRequested"
         },
         modelListeners: {
             "persistedValues.uiLanguage": {
@@ -137,7 +138,15 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
             // the storytelling tool "main" menu
             menu: {
                 type: "sjrk.storyTelling.ui.menu",
-                container: ".sjrkc-st-menu"
+                container: ".sjrkc-st-menu",
+                options: {
+                    listeners: {
+                        "onInterfaceLanguageChangeRequested.changeUiLanguage": {
+                            func: "{page}.applier.change",
+                            args: [["persistedValues", "uiLanguage"], "{arguments}.0.data"]
+                        }
+                    }
+                }
             },
             // the UIO component
             uio: {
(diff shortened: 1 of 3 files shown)


jobara:
@BlueSlug could you please try this out to see if you are still able to reproduce the race condition?

BlueSlug (approved):
These changes look good, and the tests haven't become stuck after several runs in my environment (normally it would freeze ~90% of the time). I approve 👍🏻

jobara:

src/ui/base-page.js
```diff
@@ -56,11 +56,21 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
                 target: "{that ui blockManager templateManager}.options.templateConfig.resourcePrefix"
             },
             "timeBased.stopMediaPlayerOnContextChange": {
-                record: { "{page}.events.onContextChangeRequested": "{that}.stopMediaPlayer" },
+                record: {
```
cindyli, line 59: I'm not able to reproduce the race condition locally, so I didn't trace and am not sure if the issue is caused by what's described in the JIRA.

Just by looking at the code, most of the code change is to rewrite event listeners by adding namespaces or move the same listener from the target component to the source component. If the race condition does exist, I doubt if these changes would help to fix the issue.

By looking at the test improvement, the test sequence now includes the check of more event firings such as `onReadyToBind`. This might help to deal with the race condition by ensuring the following actions are performed after an expected event has been fired. However, in my opinion, this probably hides the race condition rather than fixing it because in the real-life execution, actions in a full process won't wait in between.

Again, as I didn't successfully reproduce and track down the problem, what I said could be very wrong. I'm looking forward to hearing the rationale of how this PR fixes the race condition. Thanks.

jobara: @cindyli thanks for asking this question, it actually seems that I slightly misread some of the events. I thought the `onResourcesLoaded` event was coming from the menu, but it was actually coming from UIO. So in reality the tests were side stepping the issue in a way, as they didn't even consider the event where the tests were halting at. To address this I've pushed things out even further to wait for `onAllUiComponentsReady`. Some of the changes are cleanup around the events, but in the end I don't think they were causing the issue. I think it was an issue with the tests rather than the component. 

Below are the operations that occur when you click a menu item to change the language. I've used an ordered list to make it easier to identify which parts go together; however, the order or occurrence at the same nesting level is not guaranteed in all cases ( some use priorities ). 

_Couldn't render enough sub items in a list so I had to use a code block_

```
1. click on a menu item
    1. `menu.events.onInterfaceLanguageChangeRequested` fired
        1. `base.page` model changed (`persistedValues.uiLanguage`)
            1.`{base.page}.events.onRenderAllUiTemplates` fired
                1. `menu.templateManager.events.onResourceLoadRequested` fired
                    1. `menu.templateManager.messageLoader` created
                        1. `menu.templateManager.messageLoader.events.onResourcesLoaded` fired
                            1. `menu.templateManager.events.onMessagesLoaded` fired
                    2. `menu.templateManager.templateLoader` created
                        1. `menu.templateManager.templateLoader.events.onResourcesLoaded` fired
                            1. `menu.templateManager.events.onTemplateInjected` fired
                            2. `menu.templateManager.events.onTemplateLoaded` fired
                                1. `onTemplateLoaded` + `onMessageLoaded` combine to trigger `menu.templateManager.events.onAllResourcesLoaded`
                                    1. `menu.templateManager.renderTemplate` called
                                        1. `menu.templateManager.renderTemplateOnSelf` called
                                            1. `menu.templateManager.events.onTemplateRendered` fired
                                                1. `menu.events.onReadyToBind` fired
                                                    1. menu bind click handler to `languageLinkEnglish` element called
                                                    2. menu bind click handler to `languageLinkSpanish` element called
                                                    3. `menu.events.onControlsBound` fired ( `priority: "last"`)
                                                        1. `{base.page}.events.onAllUiComponentsReady` fired(composite event)
            2. `{base.page}.events.onContextChangeRequested` fired
                1. `stopMediaPlayer` called ( this is an invoker defined on a potential subcomponent )
            3. `cookieStore.set` called
            4. relayed to `uio.model.locale`
                1. relayed to `uio.messageLoader.model.resourceLoader.locale`
                    1. `uio.prefsEditorLoader.messageLoader.events.onCreateTOC` fired
                    2. `fluid.resourceLoader.modelUpdated` called
                        1. `uio.prefsEditorLoader.messageLoader.resourceFetcher.refetchAll` called
                            1. `uio.prefsEditorLoader.messageLoader.resourceFetcher.resourceFetcher.refetchAll` called
                                1. `fluid.fetchResources.initiateRefetch` called
                                    1. `fluid.fetchResources.initiateRefetch` called
                                        1. `fluid.initResourceFetcher` called
                                            1. `uio.prefsEditorLoader.messageLoader.resourceFetcher.onInit` fired
                                                1. `fluid.resourceLoader.subscribeCompletion` called
                                                    1. `uio.prefsEditorLoader.messageLoader.events.onResourcesLoaded` fired
                                                        1. `uio.prefsEditorLoader.events.onPrefsEditorMessagesLoaded` fired
                                                            1. `uio.prefsEditorLoader.events.onCreatePrefsEditorReady` fired (composite event)
                                                            2. `uio.prefsEditorLoader.events.onCreateSlidingPanelReady` fired (composite event)
                                                                1. `uio.prefsEditorLoader.slidingPanel` created
                                                            3. `uio.prefsEditorLoader.events.templatesAndIframeReady` fired (composite event)
                                                                1. `uio.prefsEditorLoader.prefsEditor` created
                                                                    1. `uio.prefsEditorLoader.bindReset` called
                                                        2. `uio.prefsEditorLoader.events.onCreateSlidingPanelReady` fired (seems this is fired by two different streams of events)
                                                            1. `uio.prefsEditorLoader.slidingPanel` created
                                                        3. `uio.prefsEditorLoader.events.onReady` fired (`priority: "after:recreateSlidingPanel"`)
                                                            1. `fluid.prefs.separatedPanel.bindEvents` called
                                                        4. `uio.prefsEditorLoader.events.onPrefsEditorRefresh` fired (`priority: "after:onSlidingPanelReady"`)
                                                            1. `.uio.prefsEditorLoader.iframeEnhancer.updateModel` called
                                                                1. Perform enactor updates as necessary
                                                        5. `uio.events.onUioReady.fire` fired (`priority: "after:rerenderUIO"`)
                                                            1. `{base.page}.events.onAllUiComponentsReady` fired (composite event)
```

I believe the issue with the old tests is likely due to the back-to-back `onControlsBound` and `onResourcesLoaded` event listeners. 

https://github.com/fluid-project/sjrk-story-telling/blob/d54f2a2e8383e27ae2a7de5aeeb99f11b7362b2b/tests/ui/js/base-pageTests.js#L82-L91

https://github.com/fluid-project/sjrk-story-telling/blob/d54f2a2e8383e27ae2a7de5aeeb99f11b7362b2b/tests/ui/js/base-pageTests.js#L105-L114

Both of those events start off with the model change to the `persistedValues.uiLanguage` model path in the `sjrk.storyTelling.base.page` grade. `onControlsBound` follows a path from a modelListener defined on the `sjrk.storyTelling.base.page` grade and `onResourcesLoaded` via a series of modelRelays. Both chains also involve resource fetching to get the locale specific message bundles. 

Possible issues:

1. There are variations into which of these operations completes first. In some cases the `onResourcesLoaded` event from the UIO's message loader is called before `onControlsBound`. If that happens, the event listeners may be bound in the wrong order. 
2. They could be executing in a consistent order, but the listener for `onResourcesLoaded` just happens to bind after the event has already fired. Which would seem a possibility if it really is related to machine resources, and faster machines experience the issue more readily.

I believe the IoC testing framework only looks ahead one sequence item at a time. (@amb26 would need to confirm this). If that's the case, either of the above cases can occur because the second listener hasn't been bound at the correct time.

To address the issue above, I wait for the [`onAllUiComponentsReady`](https://github.com/fluid-project/sjrk-story-telling/blob/d54f2a2e8383e27ae2a7de5aeeb99f11b7362b2b/src/ui/base-page.js#L81-L86) event which is a composite event fired after both `onControlsBound` and `onResourcesLoaded` (via `onUioReady`) have fired. In this way we know that all of the operations that were required have completed. I don't think it really matters what order these operations happen in, that is the menu and UIO being updated. It matters that they have both been updated.

amb26:
> I believe the IoC testing framework only looks ahead one sequence item at a time. (@amb26 would need to confirm this). If that's the case, either of the above cases can occur because the second listener hasn't been bound at the correct time.

Yes, that's right - it looks ahead just one sequence item. Your analysis above looks great and it looks like you might well have got to the bottom of the problem!

Note that strictly speaking any test sequence with two listeners in a row is a bug precisely because of the risk of this kind of race. https://issues.fluidproject.org/browse/FLUID-5502

cindyli:
This pull request looks good to me. Thanks for the careful explanation of the language change workflow, @jobara.

jobara:
merged at 6ed4d317be8539dbe8b61f80d3fb06d9cf8ce75b

PR #109, opened 2021-02-01
SJRK-449 Configure CI for new AWS server

gtirloni:
Please give your approval but let me merge this once the server and the configuration is ready.


BlueSlug (approved):
Assuming the new environment vars are set up, this all looks great to me. Adding `workflow_dispatch` is a good idea and gets us back to some of the functionality we had in Jenkins

PR #111, opened 2021-02-19
SJRK-402, SJRK-403 & SJRK-404: Add Signup, Login and Logout capabilities to the tool

BlueSlug:
This update will add Signup, Login & Logout controls and pages to the tool. This is currently only implemented for the `base` and `aihec` themes.

This update also includes a few bug fixes for issues that did not have Jiras filed, namely:
- the `author` data was missing on server responses for the View page
- the `createNewStoryOnServer` function was being called at page load, rather than at first user interaction as we had assumed
- (housekeeping) all `.handlebars` files have been renamed to `.hbs` to reduce filename length for better readability

Jira links:
https://issues.fluidproject.org/browse/SJRK-402
https://issues.fluidproject.org/browse/SJRK-403
https://issues.fluidproject.org/browse/SJRK-404
https://issues.fluidproject.org/browse/SJRK-442

@cindyli to review

**UPDATE:**
This update includes commented-out code which introduces the regression described by [SJRK-454](https://issues.fluidproject.org/browse/SJRK-454)


src/ui/base-page-login.js
```diff
@@ -0,0 +1,371 @@
...
+            // "ready" (the initial state), "requestSent", "responseReceived"
+            loginState: "ready",
+            loginButtonDisabled: false,
+            progressAreaVisible: false,
+            serverResponseText: false,
+            emailErrorMessage: "",
+            passwordErrorMessage: ""
+        },
+        modelListeners: {
+            progressAreaVisible: {
```
cindyli: Recently, I learnt a a simpler way to control the visibility of a dom element with the new renderer via "modelRelay" block:

```
modelRelay: {
    progressAreaVisibleChange: {
        source: "progressAreaVisible",
        target: "{loginUi}.dom.progressArea.visible"
    }
}
```

Just for your information. You don't need to make the change.

cindyli:
Clicking the "sign up" link gives:

```
isError: true
message: "Cannot GET /signUp.html"
```

Are there special configs needed for the sign up to work? I'm using "aihec" theme.

cindyli:
I didn't find the time out control for the login session. I wonder once a user logs in, will he/she remains logged in until an explicit logout?

src/ui/base-page-login.js
jobara: @cindyli thanks for this info. We've updated to the FLUID-6580 line of Infusion so I guess this should be do-able now. @BlueSlug, if you decide to do this change later, could you please file a JIRA about it.

jobara:
> Clicking the "sign up" link gives:
> 
> ```
> isError: true
> message: "Cannot GET /signUp.html"
> ```
> 
> Are there special configs needed for the sign up to work? I'm using "aihec" theme.

@cindyli I don't believe that Signup has been implemented in this PR yet. At the moment you can use Postman or other tool to send the signup request to `/signup`.

[StoryTelling.postman_collection.json.zip](https://github.com/fluid-project/sjrk-story-telling/files/6022013/StoryTelling.postman_collection.json.zip) contains my collection of Postman requests that I used to test the server. You can use the signup request from there to create an account.

src/ui/base-page-login.js
BlueSlug: I knew this was on its way, and it looks incredibly convenient! This would have applications in other parts of the codebase, too.

I'll see how quick it is to implement in this PR, and if it takes more than an hour or so I'll defer it to another Jira.

cindyli:
> > Clicking the "sign up" link gives:
> > ```
> > isError: true
> > message: "Cannot GET /signUp.html"
> > ```
> > 
> > 
> > Are there special configs needed for the sign up to work? I'm using "aihec" theme.
> 
> @cindyli I don't believe that Signup has been implemented in this PR yet. At the moment you can use Postman or other tool to send the signup request to `/signup`.

Thanks for the information, @jobara. Is the API documentation planned to be added in this PR or later?

jobara:
> Thanks for the information, @jobara. Is the API documentation planned to be added in this PR or later?

Which API docs are those? The [end-points](https://github.com/fluid-project/sjrk-story-telling/blob/main/docs/END-POINTS.md) were documented previously, if that's what you are referring to.

cindyli:
> > Thanks for the information, @jobara. Is the API documentation planned to be added in this PR or later?
> 
> Which API docs are those? The [end-points](https://github.com/fluid-project/sjrk-story-telling/blob/main/docs/END-POINTS.md) were documented previously, if that's what you are referring to.

Right. They are in the pull request that implements the server side. Now I remember. Thanks for pointing out, @jobara.

src/ui/base-page-login.js
BlueSlug: I was unable to get it working, nor was I able to find any sample code (could have sworn I saw some in another project!) or documentation. I have filed [SJRK-451](https://issues.fluidproject.org/browse/SJRK-451) to tackle this update some other time.

src/ui/base-page-storyEdit.js
```diff
@@ -35,6 +35,16 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
                     }
                 },
                 target: "{that storyEditor blockManager}.options.modelListeners"
+            },
+            "story.createNewStoryOnServer": {
+                record: {
+                    "": {
```
cindyli, line 52: With `createNewStoryOnServer` triggered by the model change to the root path of the `story` component, it means every story content change will also trigger this listener, which might be unnecessary. I wonder if it's possible to use `includeSource: "init"` to only trigger this listener at the component initialization?

src/ui/base-page.js
```diff
@@ -164,6 +171,84 @@ https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.
         }
     });
 
+    // mix-in grade to include login/logout controls on the page
+    fluid.defaults("sjrk.storyTelling.base.page.withAuthorControls", {
```
cindyli, line 180: This component is more than holding author account information. It's this plus the logout functionality. The current component structure is a bit messy in the way that:
```
sjrk.storyTelling.base.page.withAuthorControls = auth info + logout functionality
sjrk.storyTelling.base.page.login = sjrk.storyTelling.base.page.withAuthorControls + login functionality
```

The login component seems serving the entire account management with both login and logout. What I'm thinking is:
1. Rename the use of `login` for both the component name and the file name `base-page-login.js` to `accountManager`;
2. Move this component into the `accountManager` file;
3. Either create a single `accountManager` component containing login, logout, author account management functionalities, or split 3 functionalities into 3 separate components then assemble.

src/ui/ui-authorControls.js
```diff
@@ -0,0 +1,41 @@
...
+You may obtain a copy of the BSD License at
+https://raw.githubusercontent.com/fluid-project/sjrk-story-telling/main/LICENSE.txt
+*/
+
+"use strict";
+
+(function ($, fluid) {
+
+    // a UI representing the "author controls" part of every page
+    fluid.defaults("sjrk.storyTelling.ui.authorControls", {
```
cindyli, line 17: The same suggestion as above for the UI part of the account management.

cindyli:
I ran into an issue with the login page. Here are steps:

1. I followed @jobara's suggestion to manually create an account by sending a request to /signup endpoint;
2. The account appears in the CouchDB so it should have been created successfully;
3. Go to the login page, enter the email and password, hit this error:
Screen Shot 2021-02-23 at 11 56 41 AM

Did I do something wrong or miss anything? Thanks.

src/ui/base-page-storyEdit.js
BlueSlug: `init` is too early for this to fire, since that is effectively the same as having it call `createNewStoryOnServer` on page load, which is not what we're aiming for at the moment since we want to avoid clogging up the database with story records.

Currently, the function that's called by the invoker checks whether there's an `id` key on the story model, which is a workable solution in the "short" term, but ideally we would have a no-op for changes after the "first" one. It's an interesting challenge.

@jobara and @amb26 had a discussion in Matrix last week (see "Fluid Technology" from 2021-02-17), and @amb26's suggestion of using a flag to track the first model change and attaching a listener to that flag seems like a promising solution to this awkwardness:

> Just create a model listener which listens to "" and produces a {changePath: "userInteractionReceived", value: true} and then attach the listener which checks whether there is an id to that flag

cindyli: I see. Thanks for the explanation, @BlueSlug.

src/ui/base-page.js
BlueSlug: I agree that the current scheme is a bit clunky and not entirely accurate, and I was having trouble coming up with an appropriate name for this mix-in grade that didn't seem too much like a boolean variable type name like `hasAccountControls` or `isLoginEnabled`. The `ui` called `authorControls` contains the user greeting, the logout button and the link to the Login and Sign Up pages, hence why this mix-in is a page "with author controls". Those elements should be included on every page for a site that has accounts enabled.

If we implement this proposed setup as I understand it, then we would need to include the code for the login process on any page that we want to have the authorControls included on, which would increase the payload.

The structure I chose was primarily because the login elements only need to be present on the actual login page (under the current design, at least), since it is its own discrete HTML page. I feel that it might be misleading for the page that an author would go to for logging in to (and its underlying `ui` component) be called `accountManager`, as there isn't otherwise any account management happening.

cindyli: Please add this reason that results in the current component design as a code comment. Thanks.

BlueSlug:
There is currently an issue triggered by the client-side session check (in [`ui-authorControls.js`](https://github.com/BlueSlug/sjrk-story-telling/blob/17ea3f34b6443a30d81c97c011aa160f39d01bec/src/ui/ui-authorControls.js#L36-L42)) which results in the error described by [SJRK-454](https://issues.fluidproject.org/browse/SJRK-454).

BlueSlug:
Considering that whether the model listener that triggers the [SJRK-454](https://issues.fluidproject.org/browse/SJRK-454) bug is enabled or not, the author may still encounter the situation where their session has expired on the server and they cannot publish without logging on again, and considering that I will be pausing my development on this project for an indeterminate amount of time, I have decided to disable the model listener to make the tool usable again.

Ready for more review! :)

src/ui/base-page-signup.js
```diff
@@ -0,0 +1,140 @@
...
+                    },
+                    model: {
+                        confirm: null, // the new author's password confirmation
+                        confirmComparison: null // the intermediate comparison holder
+                    },
+                    modelListeners: {
+                        confirmComparison: {
+                            this: "console",
+                            method: "log",
+                            args: ["comparison from", "{change}.oldValue", "to", "{change}.value"]
```
cindyli: This debugging code can be removed.

src/ui/base-page-authentication.js
```diff
@@ -0,0 +1,361 @@
...
+            },
+            onAuthenticationRequested: "{authenticationUi}.events.onAuthenticationRequested",
+            onAuthenticationSuccess: null,
+            onAuthenticationError: null
+        },
+        listeners: {
+            // overwrite the listener to prevent losing the previous page URL
+            "onCreate.setCurrentPage": null,
+            "onAuthenticationRequested.resetEmailErrorMessage": {
+                func: "{that}.resetEmailErrorMessage",
```
cindyli: It will simplify the component definition by moving what's defined in the `resetEmailErrorMessage()` invoker directly to here. It also saves other developers from looking up the invoker level.

The same comment applies to other locations for example line 112, 125, 146 as well as the same usage in other components.

src/ui/base-page-signup.js
```diff
@@ -0,0 +1,140 @@
...
+                this: "{authenticationUi}.dom.passwordConfirmationComparisonErrorText",
+                method: "text",
+                args: ["{change}.value"],
+                priority: "after:confirmErrorMessageUpdate",
+                namespace: "confirmComparisonErrorMessageUpdate"
+            }
+        },
+        listeners: {
+            "onAuthenticationRequested.resetPasswordConfirmationErrorMessage": {
+                func: "{that}.resetPasswordConfirmationErrorMessage",
```
cindyli: The comment above about bypassing the extra layer of invokers applies to here and line 50 too.

cindyli:
> I ran into an issue with the login page. Here are steps:
> 
>     1. I followed @jobara's suggestion to manually create an account by sending a request to /signup endpoint;
> 
>     2. The account appears in the CouchDB so it should have been created successfully;
> 
>     3. Go to the login page, enter the email and password, hit this error:
> 
> 
> Screen Shot 2021-02-23 at 11 56 41 AM
> 
> Did I do something wrong or miss anything? Thanks.

It was my bad that I didn't re-install npm packages. It's gone with a fresh `npm install`.

src/ui/base-page-signup.js
BlueSlug: Whoops, good catch! I usually take care to remove those, but I obviously forgot about this one

cindyli:
Merged at e8fb396d078ee557b6ee58b7d58805ddc9d47fe1

PR #112, opened 2021-02-26
NOJIRA: fix case on login and signup page links

BlueSlug:


cindyli:
Merged at 869b48acbe2b8ee735809bfd93a8a0df3da1416f

